PR #3, opened 2017-07-28
Compare the two versions of the i429 work

MartinEden:


MartinEden:
I've manually included any changes that were important in the true `i429` branch.

PR #4, opened 2017-07-28
Add reporting API and portal to deployment

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-429

Okay, here's the final PR for this ticket - here's the actual deployment tool work.


docker-compose.yml
```diff
@@ -27,7 +36,9 @@ services:
       - api
       - admin
       - contrib
+      - report
```
littledice: does this not depend the reporting_api as well?

MartinEden: Good catch! Okay for me to merge now?

PR #5, opened 2017-08-01
I367

littledice:
Deploy script should inject a secure password into the db/api


littledice:
@MartinEden I went with local file writing and copying in the end as opposed to doing it all with docker exec and the command line, but after merging master in I noticed that you chose to add a config file for the reporting API in the latter way. So we should probably pick one method and use it (I'll defer to you on which that should be!)

src/deploy.py
```diff
@@ -24,13 +27,19 @@ def migrate_schema(db_password):
...
 
 def set_passwords_for_db_users(passwords):
-    pass
+    conn_string = "host='localhost' port='5432' dbname='montagu' user='vimc' password='changeme'"
+    conn = psycopg2.connect(conn_string)
+    cur = conn.cursor()
+    cur.execute("ALTER USER vimc WITH PASSWORD '{}'".format(passwords["api"]))
+    conn.commit()
+    cur.close()
+    conn.close()
```
MartinEden: Should there be some kind of try/finally logic around closing the cursor and connection? Or use Python's equivalent of C# using, which is `with psycopg2.connect(conn_string) as conn:`. I don't know if psycopg2 supports that, but if so, that would be the best I think.

src/service_config/api_config.py
```diff
@@ -55,3 +56,11 @@ def get_token_keypair():
...
         raise Exception("Obtaining token keypair failed: Missing file(s) in " + paths.token_keypair)
     return result
+
+
+def generate_config_file(config_path, db_password: str):
+    config_file_path = join(paths.config, "config.properties")
+    file = open(config_file_path, "w")
+    file.write("db.password={}".format(db_password))
+    file.close()
+    docker_cp(config_file_path, api_name, join(config_path, "config.properties"))
```
MartinEden: Here you can definitely use:

```
with open(config_file_path, "w") as file:
    file.write("db.password={}".format(db_password))
````

File: config/config.properties [hunk 3]
```diff
@@ -0,0 +1 @@
+db.password=WKXOQPUFO3LPDNVT7XNFQTYA7CF9EDBUXV4A551RBOAU257J0R
```
MartinEden: This file shouldn't have been committed, surely?

src/deploy.py
```diff
@@ -24,13 +27,19 @@ def migrate_schema(db_password):
...
 
 def set_passwords_for_db_users(passwords):
-    pass
+    conn_string = "host='localhost' port='5432' dbname='montagu' user='vimc' password='changeme'"
+    with psycopg2.connect(conn_string) as conn:
+        with conn.cursor() as cur:
+            cur.execute("ALTER USER vimc WITH PASSWORD '{}'".format(passwords["api"]))
+            conn.commit()
+    # unlike file objects or other resources, exiting the connection’s with block doesn’t close the connection
+    conn.close()
```
MartinEden: So should we also have a try finally here? Is it bad if you don't close the connection, or is that just fine?

littledice:
So it seems the 'with' statement does clean up the connection after all

PR #6, opened 2017-08-07
Idempotent user set up with any number of users.

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-472


MartinEden:
Rich and I tried to get the live deploy of the modified update on Friday, but we realised needed a new db user to run the update as. We [hacked something](https://github.com/vimc/montagu/commit/605c42884fe3464bf0428223bfccd663c04f1248) into place on Friday afternoon, but it didn't work properly.

This PR is a more proper treatment of the need to be able to easily add users to the database, either with passwords that are transitory (when we just want to distribute them to the containers and forget about them) or passwords from the vault for processes that need to be run at a later date (as with the modified update).

I've written it with the idea in mind that we could have a yaml file that specifies which roles each user has, but I haven't tried to do all that right now.

MartinEden:
Okay, I've made some changes:

* The new API username gets pushed into the API config
* You can choose to use testing passwords in the config. I think this is a good idea, otherwise even if you're just doing a test you need to connect to the vault, and then we're scattering the real passwords all over the place.

PR #13, opened 2017-09-04
i625 - clone and setup orderly reports properly

richfitz:
This fixes VIMC-625, VIMC-606 (and also VIMC-487 which was the old work).
   
I've checked this locally with both the "restore" and the "test_data" data sources, and it appears to work.

If we do a restore, we need to fix the permissions on ssh keys (VIMC-606).  This would be more tidily done as as post-restore hook in the database repo, which we could do by creating a subclass of `NamedVolumeTarget`, but that seems like a big hammer to hit this with.  Alternatively, we could look into why duplicati is not preserving permissions - it is meant to!

If we're doing fake data, then the code here will do a clone of the montagu-reports repo (this is different to previously where we created an empty orderly store).  I've left support in (the `clone = False` option to `configure_orderly_store`) to preserve the original behaviour - we can map that up to one of the data source options if it's relied on anywhere.  The wrinkes here are discussed in VIMC-487, but include

* to clone from a private repo we need ssh keys (recovered from the vault)
* to clone from github we need to add it to the known hosts list (the current solution may not be as good as storing known hosts in this repo perhaps)
* we need to get orderly's Postgres credentials organised into the volume for things to be useful at all


richfitz:
I see this is currently failing on teamcity due to not finding vault.  I'm not immediately seeing how to get around this.  I currently use vault unconditionally in the script and suspect that's at fault

richfitz:
This is passing on TC now

src/orderly.py
```diff
@@ -1,13 +1,103 @@
...
+    ssh = orderly_prepare_ssh(clone)
+    envir = orderly_prepare_envir(use_real_passwords)
+
+    container = "orderly_setup"
+    args = ["docker", "run", "--rm", "-d",
+            "--entrypoint", "sleep",
+            "--name", container,
+            "-v", orderly_volume_name + ":/orderly",
+            get_image_name("montagu-orderly", versions.reports),
+            "3600"]
```
MartinEden: Why 3600?

src/orderly.py
```diff
@@ -1,13 +1,103 @@
...
+        else:
+            print("creating empty orderly store")
+            run(["docker", "exec", container, "/usr/bin/orderly_init", "/orderly"],
+                check = True)
+            # NOTE: Do this _after_ init, or we can't do the
+            # initialisation as directory not empty
+            docker_cp(ssh, container, "/orderly")
+            docker_cp(envir, container, "/orderly")
+    finally:
+        paths.delete_safely(paths.orderly)
```
MartinEden: Probably better for this line to live in `deploy.py()`, where we delete other sensitive data, for consistency.

src/orderly.py
richfitz: "long enough" (1 hour).  This is a complication of not having a process running but wanting to get things into the container - I'm running `sleep 3600` to keep the container up and then stopping it later.  The number won't affect things unless it's too short

richfitz: I've added a note in cb3b791

src/orderly.py
richfitz: Fixed in 058178d

richfitz:
This is ready for another look @MartinEden

src/orderly.py
MartinEden: Why not `sleep infinity` then? If you are explicitly killing the container anyway.

richfitz: TIL `sleep infinity`

richfitz: (This is implemented now - it's apparently linux specific but as it only runs in a container that we know has a suitable bash/linux that's ok)

PR #14, opened 2017-09-06
I632 - check deploy environment is clean

richfitz:
The idea here is to prevent deploying into production a system that
is untagged and/or has a dirty git status (the latter is most
commonly triggered by a submodule being out of date - seen with the
backup module).

We only want this on production really.  We detect this by looking
to see if 'backup' is selected (this will be false for all but
production).

### DIFF
--- a/src/git.py
+++ b/src/git.py
@@ -19,6 +19,11 @@ from subprocess import run, PIPE
 # machine, so it would be good if that did not happen.
 def git_check(settings):
     strict = settings["backup"]
+    def report(msg):
+        if strict:
+            raise Exception(msg)
+        else:
+            print("NOTE: " + msg)
 
     sha = git_sha()
     if not sha:
@@ -28,11 +33,6 @@ def git_check(settings):
 
     is_clean = git_is_clean()
 
-    def report(msg):
-        if strict:
-            raise Exception(msg)
-        else:
-            print("NOTE: " + msg)
 
     if not is_clean:
         report("git status reports directory is unclean")
--- a/src/git.py
+++ b/src/git.py
@@ -45,7 +45,8 @@ def git_check(settings):
     print("This is montagu {tag} ({sha})".format(tag = tag, sha = sha))
 
 def git_is_clean():
-    p = run(["git", "status", "-s"], stdout = PIPE, stderr = PIPE)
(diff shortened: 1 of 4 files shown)


src/git.py
```diff
@@ -0,0 +1,69 @@
...
+
+    tag = git_get_tag(sha)
+    if not tag:
+        report("HEAD is not tagged")
+        tag = "<<UNTAGGED>>"
+
+    print("This is montagu {tag} ({sha})".format(tag = tag, sha = sha))
+
+def git_is_clean():
+    p = run(["git", "status", "-s"], stdout = PIPE, stderr = PIPE)
```
MartinEden: You may wish to add `check=True`, which will raise an exception if git exits with a non-zero exit code.

src/git.py
```diff
@@ -0,0 +1,69 @@
...
+    return len(p.stdout) == 0
+
+def git_get_tag(ref):
+    args = ["git", "describe", "--tags", "--exact-match", ref]
+    p = run(args, stdout = PIPE, stderr = PIPE)
```
MartinEden: Similarly `check=True` here.

src/git.py
```diff
@@ -0,0 +1,69 @@
...
+        tag = p.stdout.decode("utf-8").strip()
+    else:
+        tag = None
+    return tag
+
+def git_sha():
+    args = ["git", "rev-parse", "HEAD"]
+    p = run(args, stdout = PIPE, stderr = PIPE)
+    code = p.returncode
+    if code == 128:
```
MartinEden: A comment to explain this magic number would be good.

src/git.py
richfitz: For some reason I thought that for unclean status git returned exit code > 0 but that is not actually the case.  I've added this here

src/git.py
richfitz: ...but down here I check the return code explicitly in the next line because `describe tags` does return nonzero if tags aren't found

MartinEden: You're right, sorry!

src/git.py
richfitz: yeah, agreed

richfitz:
This is ready for another look @MartinEden

PR #17, opened 2017-09-18
Use root user for migrations.

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-686


MartinEden:
Ran into a problem where the migrator user didn't own the tables, so it couldn't ALTER them.

PR #19, opened 2017-09-22
Option to add a test user with access to all modelling groups

littledice:


src/cli.py
```diff
@@ -43,6 +43,27 @@ def cli():
...
+    name = get_image_name("montagu-cli", versions.api)
+    args = ["add", "Test User", "test.user", "test.user@imperial.ac.uk", "password"]
+
+    run(command + [name] + args)
+
+    args = ["addRole", "test.user", "user"]
+
+    run(command + [name] + args)
+
+    args = ["addUserToGroup", "test.user", "ALL"]
```
richfitz: Can we add them as a reports reviewer too?

`args = ["addRole" "test.user" "reports-reviewer"]`

should be enough.  Then they get access to the reporting portal too

richfitz:
Also this is failing on teamcity: http://teamcity.montagu.dide.ic.ac.uk:8111/viewLog.html?buildTypeId=montagu_deploy&buildId=5274

You need to add the new configuration option into the teamcity-settings.json file (at the repo root)

richfitz:
Looks like it's passing now :tada:

PR #28, opened 2017-10-09
Only do backup setup if required, removing need for Vault in some cases

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-769


src/backup.py
```diff
@@ -24,13 +24,15 @@ def configure(settings):
         f.write(config)
 
 
+def needs_setup():
+    return run("../backup/needs-setup.sh", stdout=DEVNULL, stderr=DEVNULL).returncode == 1
+
```
littledice: I can't find the file `needs-setup.sh` in the submodule, is it missing a commit?

PR #29, opened 2017-10-16
i810

richfitz:


src/versions.py
```diff
@@ -1,9 +1,9 @@
...
-db = "399a5d3"
-contrib_portal = "5cf035b"
-admin_portal = "5cf035b"
-report_portal = "5cf035b"
+api = "1486e06"
+reporting_api = "d8452af"
+db = "3f3d0d6"
+contrib_portal = "50d88be"
+admin_portal = "50d88be"
+report_portal = "50d88be"
```
littledice: not sure why all these versions have changed? maybe revert them to keep this update confined to the addition of orderly server

richfitz: I think it's because I need the recent version of reporting_api and dragged in updated versions of everything else - I'll revert those and see if it all works

richfitz:
Sure - the conflicts are just the versions.py file

PR #30, opened 2017-10-18
I818

richfitz:
As currently set up, the git ssh bits are not properly set up on uat


richfitz:
actually, I think it makes sense to roll VIMC-819 into this too

richfitz:
OK, I think this is now actually ready for review

PR #32, opened 2017-10-26
Release script

MartinEden:
This doesn't do everything yet, but it's something!

In particular, I'd like it if it could track which branches have been merged in the dependent repos (API, webapps, etc.), and I'd like it if it could automatically mark tickets as deployed.


scripts/release/release.py
```diff
@@ -0,0 +1,103 @@
...
+        branches_and_tickets = check_tickets(latest_tag)
+        new_tag = get_new_tag()
+
+        print("Writing release log...")
+        release_message = make_release_message(new_tag, branches_and_tickets)
+        write_release_log(release_message)
+        commit_tag_and_push()
+
+        print("Done!")
+        print("""The released tickets should now be moved into the 'Deployed'
```
littledice: Technically this isn't true yet! This script will tag and push tag to git but not actually deploy the code... presumably the flow is run this on UAT, deploy to UAT, then post user acceptance testing deploy to prod? 

Could you update the README to explain the flow?

MartinEden: Yeah, I wasn't sure about this. Maybe we want an extra state in YouTrack which is "Deployed to UAT". Then you run this script and it moves the tickets from "Ready to deploy" to "Deployed to UAT" (or, for now, prompts you to do that manually). Then we have another script that just moves the tickets from "Deployed to UAT" to "Deployed"?

littledice: I think it only make sense for tickets to be automatically moved to 'deployed' (or 'deployed to UAT') by a script that does actually deploy code.

Really we want a pre-deploy step which is tagging the release (as done here), a deploy step (as we already have), and a post-deploy step, which is updating the tickets.

I think the flow we're aiming for ultimately is something like:

* deploy to UAT happens automatically on a successful Montagu build in Teamcity 
* post-deploy step on UAT moves tickets to 'Deployed to UAT' state
* manual user acceptance testing takes place
* release is created via manual intervention, i.e. running this script on UAT, which would then look at the tickets in 'deployed to UAT' state and include those in the release log
* deploy to prod (maybe `./src/deploy.py' should take a tag as command line argument here and automatically fetch that first)
* post-deploy step on prod moves tickets to 'Deployed' state

littledice: Probably we should be thinking about external tools that can manage this flow. But either way I think for now just remove this prompt about tickets, and update the README so it's clear that this script creates a release, rather than deploying a release... 

Maybe the release log should actually have a flag to say whether a particular release was ever deployed (and when), set to false here and updated by the deploy script at the end of a successful deploy?

MartinEden: Okay, for now I've just made it clear the script doesn't actually release anything. It's now called `make-release.py` and it doesn't tell you to update the tickets.

scripts/release/helpers.py
```diff
@@ -0,0 +1,9 @@
+import subprocess
+import shlex
+
+
+def run(cmd):
+    parts = shlex.split(cmd)
```
richfitz: I'm confused why this is needed and not `*args` - shell quoting is nasty enough passing args though seems tidier to me (but there may well be reasons here that I'm missing)

scripts/release/tickets.py
```diff
@@ -0,0 +1,98 @@
+import os
+import re
+import requests
+
+from branch_diff import get_branch_diff
+
+branch_pattern = re.compile(r"^i(\d+)($|_)")
```
richfitz: Consider `"^i(\d+)($|[_-])"` - I think there's a few cases of that

scripts/release/tickets.py
```diff
@@ -0,0 +1,98 @@
...
+NOT_FOUND = "NOT FOUND"
+
+
+def get_token():
+    token = os.environ.get('YOUTRACK_TOKEN')
+    if not token:
+        print("""Please set your YouTrack token as follows:
```
richfitz: this is great

scripts/release/helpers.py
MartinEden: I just like being able to write:

```
run("git push --tags")
```

as opposed to

```
run(["git", "push", "--tags"])
```

Of course, we can always drop back to the underlying `subprocess.run` method for more complicated cases, but for the simpler cases I think this is more readable.

scripts/release/tickets.py
MartinEden: Good idea. Fixed.

MartinEden:
I've add documentation (shaky as it is) for the release process.

PR #35, opened 2017-11-07
i823 - New restore script that fixes db permissions

richfitz:
Tested locally.  Once deployed I will update the (currently disabled) cron script and that will re-enable nightly restores on science

### DIFF
--- a/src/database.py
+++ b/src/database.py
@@ -154,7 +154,7 @@ def for_each_user(root_password, users, operation):
         conn.commit()
 
 
-def setup(use_real_passwords):
+def setup(use_real_passwords, migrate = True):
     print("Setting up database users")
     print("- Scrambling root password")
     if use_real_passwords:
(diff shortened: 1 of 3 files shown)


src/database.py
```diff
@@ -154,7 +154,7 @@ def for_each_user(root_password, users, operation):
         conn.commit()
 
 
-def setup(use_real_passwords):
+def setup(use_real_passwords, migrate = True):
```
littledice: as discussed, it should always be fine to apply migrations

src/restore_db.py
```diff
@@ -0,0 +1,21 @@
...
+import backup
+from service import service
+from settings import get_settings
+from os.path import abspath, dirname
+from os import chdir
+
+def restore_db():
+    ok = service.status == 'running' and service.volume_present
+    if not ok:
+        raise Exception('montagu not in a state we can restore')
```
MartinEden: It would be good if this could print out the current services status and volume present values, so that we can debug from looking at logs later.

MartinEden: I've raised this as https://vimc.myjetbrains.com/youtrack/issue/VIMC-957 since the PR is already merged.

PR #36, opened 2017-11-07
i587 - deploy support for machine-scoped database passwords

richfitz:
A summary of changes (duplicated from 848e478d773492da19ef102ab3eee376899e0a10):

* The configuration option use_real_passwords (bool) is replaced by
  an enumeration "password_group".  The password group "fake" is
  mapped to None and represents the use_real_passwords=False option,
  with the True option being any other valid enum value
* Sending emails now only happens if password_group="production".
  We can revert to the previous behaviour by chosing password_group
  is not None
* The VaultPassword class picks up a machine_group argument and
  through this now handles the fake password case too.  This removes
  quite a few conditionals from within functions

I'm using 'password_group' here rather than 'machine' to separate
passwords (cf VIMC-587) because one might use a single password_group
on multiple machines.


littledice:
Have merged i823, you'll need to update as that branch introduces `use_real_passwords` in another place...

src/deploy.py
```diff
@@ -77,13 +77,13 @@ def configure_montagu(is_first_time, settings):
...
 
-    passwords = database.setup(settings["use_real_passwords"])
+    passwords = database.setup(settings["password_group"])
 
     # Push secrets into containers
     cert_paths = get_ssl_certificate(settings["certificate"])
     token_keypair_paths = get_token_keypair()
 
-    configure_api(passwords['api'], token_keypair_paths, settings["hostname"], settings["use_real_passwords"])
+    configure_api(passwords['api'], token_keypair_paths, settings["hostname"], settings["password_group"] == 'production')
```
MartinEden: This code would be clearer if we put that last argument in a variable - just so it's clear what its purpose is. e.g.

```
send_emails = settings["password_group"] == 'production'
configure_api(passwords['api'], token_keypair_paths, settings["hostname"], send_emails)
```

richfitz:
Can you take a look at this again when you get a moment @MartinEden?  The shell password script has now gone to be replaced by [some python](https://github.com/vimc/montagu/blob/i587/src/generate_passwords).

As I mentioned in your office I have removed mention of `scripts/get_db_password.sh` from the README (replaced with something that will work with the new layout) and propose removing that shell script entirely

README.md
```diff
@@ -73,32 +73,53 @@ However, care should be taken here because if the schema has been migrated in th
 See [here](docs/DisasterRecovery.md)
 
 ## Passwords
+
+Database passwords are managed by the vault and laid out ascii
```
MartinEden: Why does this say "ascii"?

README.md
```diff
@@ -73,32 +73,53 @@ However, care should be taken here because if the schema has been migrated in th
...
 
-This gets the password for the `readonly` user. To see all users, use
-`vault list secret/database/users`
+This gets the password for the `readonly` user for the `science` password group (which is suitable for the database running on support.montagu.dide.ic.ac.uk). To see all users, use
 
-### Generate
+```
+vault list secret/database/science/users
+vault list secret/database/production/users
+```
```
MartinEden: What's the point of this code block? Maybe add a note to explain.

README.md
```diff
@@ -73,32 +73,53 @@ However, care should be taken here because if the schema has been migrated in th
...
+
+### Add a new user
```
MartinEden: Make it clearer that we're talking about database users not Montagu users here, since this is the main Montagu docs.

README.md
```diff
@@ -73,32 +73,53 @@ However, care should be taken here because if the schema has been migrated in th
...
-./scripts/generate_passwords.sh
+./src/generate_passwords.py add_user <username>
 ```
 
-At present this manages only the database passwords, with the layout of
+Passwords can be regenerate by running the script
 
 ```
-/secret/database/:machine/users/:username
+./src/generate_passwords.py <group>
```
MartinEden: Missing an argument to the script here, I think

README.md
richfitz: Emacs has an out of control autocompletion feature that occasionally does this

README.md
richfitz: This is the code corresponding to the text above ("to see all users, use").  I've spaced it out a bit better now

README.md
richfitz: thanks

PR #38, opened 2017-11-08
i805 - Send orderly container an interrupt before stopping

richfitz:
As documented in VIMC-805, the orderly container will respond quickly
to an interrupt, but not to whatever docker stop (via docker-compose stop)
is sending.  This is (presumably) a limitation of httpuv and not something
I can see how to work around at the R level.  So instead we send an
interrupt signal (SIGINT) just before the stop, and that seems to bring
things down much more quicky.

### DIFF
--- a/src/service.py
+++ b/src/service.py
@@ -77,6 +77,10 @@ class MontaguService:
         return self._get(proxy_portal_name)
 
     @property
+    def orderly(self):
+        return self._get(orderly_name)
+
+    @property
     def volume_present(self):
         return volume_name in [v.name for v in self.client.volumes.list()]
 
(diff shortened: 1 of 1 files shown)


src/service.py
```diff
@@ -85,6 +89,8 @@ def _get(self, name):
 
     def stop(self, settings):
         print("Stopping Montagu...", flush=True)
+        if self.orderly:
+            self.orderly.kill("SIGINT")
```
MartinEden: Please add a comment here mirroring the one in PR description, to explain why we are doing this.

richfitz: done

PR #40, opened 2017-11-10
Replace versions.py with submodules

MartinEden:
You will need to run `git submodule init` and `git submodule update` after updating to a version of the repo that has these submodules.


.gitmodules
```diff
@@ -1,6 +1,30 @@
 [submodule "backup"]
 	path = backup
 	url = https://github.com/vimc/montagu-backup
-[submodule "database"]
-	path = database
-	url = https://github.com/vimc/montagu-db.git
```
MartinEden: We added this database submodule at one point when we were running migrations manually.

teamcity.sh
```diff
@@ -1,7 +1,6 @@
...
 set -e
 cp teamcity-settings.json src/montagu-deploy.json
-cd src
-pip3 install -r requirements.txt
-./deploy.py
-./test.py $@
\ No newline at end of file
+pip3 install --quiet -r src/requirements.txt
+./src/deploy.py
+./src/test.py $@
```
MartinEden: Need to run `deploy.py` as `./src/deploy.py` or it can't find the submodules. Also, this is how we always run it for deployments, so TeamCity should be doing the same thing as us!

src/versions.py
```diff
@@ -1,12 +1,44 @@
...
-report_portal = "56df438"
-cert_tool = "ae10ceb"
-proxy = "623c7be"
-orderly = "c307429"
+import re
+from os.path import join
+from subprocess import run, PIPE
+
+
+def get_submodule_version(path):
```
richfitz: A possibly simpler way of achiving this:

```
git -C backup rev-parse --short HEAD <path>
```

will return the short hash for the path

src/update_versions_to_latest_master.py
```diff
@@ -1,37 +1,18 @@
...
-        ("db", "montagu-db"),
-        ("contrib_portal", "montagu-webapps"),
-        ("admin_portal", "montagu-webapps"),
-        ("report_portal", "montagu-webapps"),
-        ("cert_tool", "montagu-cert-tool"),
-        ("proxy", "montagu-proxy"),
-        ("orderly", "montagu-orderly")
-    ]
+    root = "submodules"
+    for submodule in os.listdir(root):
```
richfitz: Possibly simpler way of achieving this is to use:

```
git submodule update --recursive --remote
```

MartinEden: Ah yes, that's much nicer.

MartinEden:
I looked into using bare submodules, because of course we don't actually need the code checked out, but I couldn't find any documentation for doing that. Do you know how, @richfitz?

src/versions.py
richfitz: Sorry this should be `git -C <path> rev-parse --short HEAD`

richfitz:
bare submodules aren't possible so far as I can see.  So if we go the bare route we're on our own with cloning and updating.  But we're also not needing anything like the power and flexibility of submodules at that point

src/versions.py
MartinEden: Again, a much better way. Clearly Googling "how to do X in Git" is not the way forwards!

MartinEden:
I think for now, let's stick with the submodules, as we have that working, and other than the portals (which will improve once we strip the templates out) the submodules are all tiny.

littledice:
Nice, think this is a good way forward!

PR #45, opened 2017-11-15
Make release tool work with submodules

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-974

The easiest way to test this out is to checkout this branch, then run `./src/update_versions_to_latest_master.py` to update the submodules, then run the `./scripts/release/make_release.py`. Then you'll get lots of nice juicy changes being reported. Just don't forget to revert the submodules afterwards!


scripts/release/branch_diff.py
```diff
@@ -11,18 +15,66 @@ def __init__(self, remote, name):
...
         if " -> " in raw:
             return None
         else:
-            parts = raw.split('/')
-            return Branch(parts[0], parts[1])
+            split_point = raw.find('/')
+            if split_point:
+                return Branch(raw[:split_point], raw[(split_point + 1):])
+            else:
+                raise Exception("Unable to parse branch '{}'".format(raw))
```
MartinEden: Turns out we had a branch name in the montagu-webapps with a `/` in its name, so this needed adapting.

scripts/release/git_helpers.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+def get_submodule_version(path):
+    full_path = join("submodules", path)
+    result = run(["git", "-C", full_path, "rev-parse", "--short=7", "HEAD"],
+                 stdout=PIPE, check=True, universal_newlines=True)
+    return result.stdout.strip()
+
+
+def get_past_submodule_version(path, master_repo_version):
```
MartinEden: I suspect there's a simpler way of doing this - there has been every time so far! Would appreciate your input, @richfitz.

scripts/release/make-release.py
```diff
@@ -71,9 +64,8 @@ def write_release_log(message):
 def commit_tag_and_push():
     run("git add RELEASE_LOG.md")
     run("git commit -m \"{msg}\"".format(msg=release_message))
-    print("Tagging and pushing...")
+    print("Tagging...")
     tag(new_tag, release_message)
-    push()
```
MartinEden: Alex and I decided it would be better if the tool auto-committed, but didn't push, so you have a last minute chance to review and amend if need be.

scripts/release/make-release.py
```diff
@@ -7,9 +7,7 @@
 from tickets import check_tickets
 
 release_tag_pattern = re.compile(r"^v\d\.\d\.\d(-RC\d)?$")
-# Enable this in development to avoid pushing
-dry_run = True
-
+dry_run = False
```
richfitz: Can you make this a command line option?  We already use docopt in `src/generate_passwords.py` so that should make this not a big faff

scripts/release/make-release.py
richfitz: :+1: this seems a better idea

scripts/release/branch_diff.py
```diff
@@ -11,18 +15,66 @@ def __init__(self, remote, name):
...
+        current_version = get_submodule_version(submodule)
+        previous_version = get_past_submodule_version(submodule,
+                                                      past_main_repo_version)
+        d = Difference.get_branch_diff(current_version, previous_version,
+                                       working_dir=full_path)
+        self.add(d, submodule)
+
+    @classmethod
+    def get_branches_at(cls, revision, working_dir=None):
+        raw = run("git branch -r --merged {}".format(revision),
```
richfitz: This seems like the only place where the `working_dir` option is used - perhaps simpler would be to use

```
run("git -C {} branch -r --merged {}".format(working_dir, revision))
```

since we use that directory changing approach elsewhere already (and then dropping that argument to run)

scripts/release/git_helpers.py
richfitz: Apparently this is the right way of doing it (you're trying to find the SHA for a submodule when the master was at `master_repo_version`, right?).  Odd that it's such an ugly command

scripts/release/make-release.py
```diff
@@ -97,6 +89,10 @@ def fetch():
         write_release_log(release_message)
         commit_tag_and_push()
 
-        print("Done!")
-        print("""When you come to deploy this release, the RELEASE_LOG.md file
+        print("""Done"
+No changes have been pushed, so please review and then push using 
+
+git push --tags
```
richfitz: Does this not push _just_ the tags?  Apparently you can do `git push --follow-tags` and it'll push the tags referenced on pushed commits (https://stackoverflow.com/questions/3745135/push-git-commits-tags-simultaneously)

scripts/release/branch_diff.py
richfitz: Does this mean that we will choke on any branchname that is not correctly specified (thinking in particular of the xxx-patch-1 branches that direct editing on github generate - like Kim might end up doing)

MartinEden: What we are doing here is splitting a remote branch's name. So this is taking something in the form `origin/my_branch` and making sure we get the `my_branch` bit. I think all remote branches will follow this format?

scripts/release/git_helpers.py
MartinEden: :+1:

scripts/release/branch_diff.py
MartinEden: I had a go at this, but it's a bit nasty. Because `working_dir` can be `None`, so then you have to conditionally include the `-C`. Whereas the current working dir argument to `subprocess.run` does the right thing when you pass it `None` for free.

scripts/release/make-release.py
MartinEden: Good idea, done.

scripts/release/make-release.py
```diff
@@ -1,14 +1,26 @@
...
+"""
+Tags a release and writes out a changelog, first checking that associated
+tickets are in the correct status. The changelog is written to RELEASE_LOG.md
+and includes ticket summaries from YouTrack, where applicable
+
+Usage:
+  make-release.py [--test-run]
+
+Options:
+  --test-run    Don't insist on Git being in a clean state
```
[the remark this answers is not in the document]
MartinEden: It's useful when you're developing the release tool, as you can test the release tool without having to commit the changes you have made to it.

PR #46, opened 2017-11-15
i880 - deploy work for the annex

richfitz:
This goes along with https://github.com/vimc/montagu-db/pull/26

* For testing I'm adding an extra container to the constellation - this is not used in production.
* One new enum field: `db_annex_type` with values
  - `fake`: for testing - use another container for the annex but on the same physical machine.  Migrate the schema but start otherwise empty.  Passwords are changeme (vimc) and changeme2 (readonly)
  - `real`: for production - use the annex machine for the annex.  Migrate the schema.  Passwords come from the vault (`/secret/annex/users/vimc` and `/secret/annex/users/readonly`)
  - `readonly`: as for `real` but don't migrate the schema for the annex

The table `burden_estimate_stochastic` in the database is essentially readonly.  All the montagu-db users map to the `readonly` user.  This is because `COPY` won't work across databases so in the api we will need a second db connection to do the bulk data import.  Some additional care will be needed on staging machines to test stochastic estimate uploading as a result.


src/database.py
```diff
@@ -136,18 +149,64 @@ def set_permissions(db, user):
...
-        image,
-        "migrate", "-user=" + root_user, "-password=" + db_password
-    ], check=True)
-
+    cmd = ["docker", "run", "--rm", "--network=" + network_name, image]
+    if host:
+        cmd += ["-url=jdbc:postgresql://{}/montagu".format(host)]
+    if config_file:
+        cmd += ["-configFile=" + config_file]
+    if placeholders:
```
littledice: my feeling is that `migrate_schema_core` and `migrate_schema_annex` are sufficiently different that there's no point in them both calling through to the same function like this - it's not very easy to see what each is doing

src/compose.py
```diff
@@ -3,24 +3,33 @@
...
+    run("pull", port, hostname, False)
+
+
+def run(args, port, hostname, fake_db_annex):
+    prefix = 'docker-compose --project-name montagu '
+    if fake_db_annex:
+        # NOTE: it's surprising that the '../' is needed here, but
+        # docker-compose apparently looks like git through parent
+        # directories until it finds a docker-compose file!
+        prefix += '-f ../docker-compose.yml -f ../docker-compose-annex.yml '
```
littledice: so if we're faking the annex, we include a container for the annex, but otherwise we don't? I'm a little confused by this

richfitz: You have it exactly right :)

src/service.py
```diff
@@ -29,18 +31,30 @@
...
+#
+# However, this is actually really hard to do correctly - even passing
+# 'settings' through as an argument to status() is hard because it's
+# used to _load_ the settings and confirm that all settings are filled
+# out appropriately!
 class MontaguService:
     def __init__(self):
         self.client = docker.from_env()
 
-    @property
```
richfitz: revert back to being a property and drop the settings check here - we'll move from a global status to one parameterised by the settings in a separate PR

src/database.py
```diff
@@ -184,15 +246,27 @@ def setup(password_group):
         print(" - {name}: {source}".format(name=user.name, source=user.password_source))
         passwords[user.name] = user.password
 
+    # NOTE: As I work through this - why not set up users *after* the
```
richfitz: the reason is that the migrator user needs to exist before doing the migrations

PR #47, opened 2017-11-20
i1052 - Post in slack as deploy begins/ends

richfitz:
This will hopefully reduce the number of "is montagu being deployed" or "why is my connection not working" questions.  Uses slack incoming webhooks, with the address stored in the vault.

This adds two more settings (I don't think they can sensibly be combined into one) - the channel to post into and the name of the deployment.

To test, just set these as something like

```
    "notify_channel": "montagu-deploy",
    "instance_name": "local (rich)",
```

and on deploy it will print a message as deployment starts, previous versions are brought down and on completion or error.  An example is in #montagu-deploy now

### DIFF
--- a/src/setting_definitions/__init__.py
+++ b/src/setting_definitions/__init__.py
@@ -85,6 +85,15 @@ definitions = [
                       "Otherwise, just use the default",
                       default_value="https://support.montagu.dide.ic.ac.uk:8200",
                       is_required=vault_required),
+
+    SettingDefinition("notify_channel",
+                      "What slack channel should we post in?",
+                      "e.g., montagu. Leave as the empty string to not post",
+                      default_value=""),
+    SettingDefinition("notify_instance_name",
+                      "What is the name of this instance to post in a channel?",
+                      default_value="(unknown)"),
+
     BooleanSettingDefinition("clone_reports",
                              "Should montagu-reports be cloned?",
                              "If you answer yes, then we need vault access in order to get the ssh keys for vimc-robot "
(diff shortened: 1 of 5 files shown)


src/setting_definitions/__init__.py
```diff
@@ -16,6 +16,7 @@ def vault_required(settings):
            or settings["certificate"] == "production" \
            or settings["certificate"] == "support" \
            or uses_vault_passwords \
+           or settings["notify_channel"] != "" \
```
MartinEden: More pythonic to just write `or settings["notify_channel"]`?

src/setting_definitions/__init__.py
```diff
@@ -85,6 +86,15 @@ def vault_required(settings):
                       "Otherwise, just use the default",
                       default_value="https://support.montagu.dide.ic.ac.uk:8200",
                       is_required=vault_required),
+
+    SettingDefinition("notify_channel",
+                      "What slack channel should we post in?",
+                      "e.g., montagu. Leave as the empty string to not post",
+                      default_value=""),
+    SettingDefinition("notify_instance_name",
```
MartinEden: Maybe we should just have an `instance_name` setting. Might turn out to be useful elsewhere?

src/setting_definitions/__init__.py
richfitz: Pythons truthyness behaviour is so weird.  But you're right - I'll do that

src/setting_definitions/__init__.py
richfitz: For some reason I thought that you wanted to avoid that (or at least a "production" flag).  But I can see how it might come in useful elsewhere (e.g., printed in the portals perhaps)

richfitz:
> I can't see the #montagu-deploy channel

You should just be able to join it - it's public.  I'll invite people in as this is information everyone should be able to see

PR #49, opened 2017-11-22
Allow one button deploy of a tag

MartinEden:
Whichever of you has time to review this, please do :)

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 set -e
 git fetch --all
-latest=$(git tag -l 'v*' | sort | tail -n 1)
+latest=$(git tag -l 'v*' | sort -n | tail -n 1)
 echo -n "Which tag do you want to deploy? [$latest] "
 read version
 version=${version:-$latest}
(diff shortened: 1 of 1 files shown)


deploy.sh
```diff
@@ -1,4 +1,14 @@
 #!/usr/bin/env bash
 set -e
-pip3 install -r src/requirements.txt
+git fetch --all
+latest=$(git tag -l 'v*' | sort | tail -n 1)
```
richfitz: A suggestion here:

```
versions=$(git tag --list 'v*' | sort -n | tail -n5 | head -n4 | tr '\n' ' ')
latest=$(git tag -l 'v*' | sort -n | tail -n 1)
echo -n "Which tag do you want to deploy? ...$versions[$latest] "
```

in any case, `sort -n` is useful if we hit double-digits on a tag part as it should sort better (11 after 2)

richfitz: The above will print

```
Which tag do you want to deploy? ...v0.7.1 v0.7.2 v0.7.3 v0.7.4 [v0.7.5] 
```

richfitz:
(the only part of that you should definitely change is using `sort -n` over `sort`)

MartinEden:
I've made the minimal change - I'd like to get this merged in and then maybe extend it further, if that's okay?

PR #50, opened 2017-11-23
i1075: Automatically tag tickets as part of release 

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-1075

### DIFF
--- a/scripts/release/make-release.py
+++ b/scripts/release/make-release.py
@@ -88,10 +88,26 @@ def fetch():
     run("git fetch --tags --all")
 
 
+def update_youtrack(branches_and_tickets, test_run):
+    if test_run:
+        print("--test-run is enabled: Skipping updating YouTrack")
+        return
+
+    print("Updating YouTrack...")
+    tickets = list(ticket for branch, ticket in branches_and_tickets)
+    problems = tag_tickets(tickets, new_tag)
+    if problems:
+        print("Error updating YouTrack:")
+        for problem in problems:
+            print(problem)
+        print("")
+
+
 if __name__ == "__main__":
     args = docopt(__doc__)
+    test_run = args["--test-run"]
 
-    if not (git_is_clean() or args["--test-run"]):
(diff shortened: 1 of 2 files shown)


scripts/release/make-release.py
```diff
@@ -104,9 +104,19 @@ def fetch():
         print("Writing release log...")
         release_message = make_release_message(new_tag, branches_and_tickets)
         write_release_log(release_message)
-        commit_tag_and_push()
+        commit_and_tag()
+
+        print("Updating YouTrack...")
+        tickets = list(ticket for branch, ticket in branches_and_tickets)
+        problems = tag_tickets(tickets, new_tag)
```
richfitz: I think this should not be done when `--test-run` is passed in (otherwise one can't do the "see what would be released").  If we do update the tags then it should ideally be done at the same time as a push so that the two don't get out of sync.  Otherwise if you make a release but don't go through and push it (but undo that commit to back out) the tickets are going to be in an inconsistent state.

Or I could have the logic wrong

MartinEden: Agreed about `--test-run`, I'll fix that.

But yeah, this is tricky, because there's no way to make the change atomic across YouTrack. I figured that we didn't want to auto-push to GitHub, and we didn't want to have to manually run another step to update YouTrack.

This way, if you change your mind, you just need to find the tickets in YouTrack with the new tag, and revert them back to being empty, which is pretty simple. We could add another script to do that reversion for you, if it came up often.

richfitz: Adding test run seems enough for now - we can iterate over this until we work out something that works

MartinEden: Fixed `--test-run`.

richfitz:
One other thought - does the message at the end of make-release need amending too if the script is updating youtrack issues?

MartinEden:
I've updated the final release message, and also generally tidied up the output.

PR #51, opened 2017-11-24
i979 - deploy from docker hub

richfitz:
This PR does a couple of related things:

* `scripts/release/tag-images.py` coordinates tagging images that were used as part of a release and publishing those onto docker hub
* changes to allow deployment using these issues (this removes almost all of our `docker.montagu.dide.ic.ac.uk:5000` strings throughout the code

The idea is we'll have all images that were part of a release lodged on docker hub, tagged with both version numbers and with the git sha.  Eventually we might want to allow using the versioned tags in deployment too (rather than the sha).

In order to push to docker hub you will need to make an account and then have me add you to the vimc account. Anyone can pull the images.

Commit 19eb21d4102049c62ae3bc6392ed8a1b11458d4a introduces changes to the way that the release script handles version numbers - the log entry for that commit details most of this - @MartinEden please check this work in particular.

To try a release from docker hub do

```
MONTAGU_USE_DOCKER_HUB=true ./src/deploy.py
```

(see the comments in src/docker_helpers.py about why this particular bit of ugliness seems like the least-pain way to put this together right now and hopefully we can come up with something better)

### DIFF
--- a/scripts/release/tag-images.py
+++ b/scripts/release/tag-images.py
@@ -41,14 +41,15 @@ container_repo_map = {
     "montagu-reverse-proxy": "proxy"
 }
 
-registry = "docker.montagu.dide.ic.ac.uk:5000"
+registry_local = "docker.montagu.dide.ic.ac.uk:5000"
+registry_hub = "vimc"
 
 def set_image_tag(name, version):
     repo_name = container_repo_map[name]
     sha = git_helpers.get_past_submodule_version(repo_name, version)
     d = docker.client.from_env()
-    img = d.images.pull("{}/{}:{}".format(registry, name, sha))
-    tag_and_push(img, registry, name, version)
+    img = d.images.pull("{}/{}:{}".format(registry_local, name, sha))
+    tag_and_push(img, registry_local, name, version)
 
 def set_image_tags(version):
     print("Setting image tags")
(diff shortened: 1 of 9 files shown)


scripts/release/tag-images.py
```diff
@@ -0,0 +1,97 @@
...
+    d = docker.client.from_env()
+    print("Pushing release to docker hub")
+    for name in container_repo_map.keys():
+        img = d.images.get("{}/{}:{}".format(registry_local, name, version))
+        publish_image(img, name)
+
+def publish_image(img, name):
+    tags = img.tags
+    published = [x.split(":")[-1] for x in tags if x.startswith(registry_hub)]
+    existing = [x.split(":")[-1] for x in tags if x.startswith(registry_local)]
```
MartinEden: I feel like this could be more readable. Maybe replace `x.split(":")[-1]` with a named function (`get_version_from_docker_tag`) or even a class. e.g.

```
class DockerTag:
    def __init__(self, raw):
        parts = raw.split("/")
        self.registry = parts[0]
        parts = parts[1].split(":")
        self.name = parts[0]
        self.version = parts[1]

...

published = [t.version for t in [DockerTag(x) for x in tags] if t.registry == registry_hub]
```

richfitz:
Try running

```
MONTAGU_USE_DOCKER_HUB=true ./src/deploy.py
```

scripts/release/tag-images.py
richfitz: I've gone with the class approach and also eliminated some repetitive formatting operations

PR #52, opened 2017-11-28
I1067

richfitz:


staging/shared/restore/restore.sh
```diff
@@ -0,0 +1,34 @@
...
+
+MONTAGU_PATH=montagu
+
+if [ -f $MONTAGU_PATH ]; then
+    (cd $MONTAGU_PATH && ./src/stop.py)
+    rm -rf $MONTAGU_PATH
+fi
+
+git clone --recursive https://github.com/vimc/montagu $MONTAGU_PATH
+git -C $MONTAGU_PATH checkout i1067
```
richfitz: both these two lines can come out

src/backup.py
```diff
@@ -53,5 +54,9 @@ def schedule(settings):
 def restore(settings):
     print("Restoring from remote backup")
     setup(settings)
+    ## Because of the annex work we need to ensure that users *exist*
+    ## here at this point and do that just before removing the
```
littledice: should this say 'before removing' or 'before restoring' the database?

PR #54, opened 2017-11-28
i1096 - fix db restore without redeploy on science

richfitz:
We will *eventually* get this running as a cron job.

The changes in versions.py make things a little less sensitive to working directories.  The other changes are mostly about getting the notifier working and creating test users if required

### DIFF
--- a/src/restore_db.py
+++ b/src/restore_db.py
@@ -1,19 +1,34 @@
 #!/usr/bin/env python3
 
-import database
 import backup
+import database
+from notify import Notifier
 from service import service
 from settings import get_settings
 from os.path import abspath, dirname
 from os import chdir
+from cli import add_test_user
 
 def restore_db():
-    ok = service.status == 'running' and service.volume_present
-    if not ok:
-        raise Exception('montagu not in a state we can restore')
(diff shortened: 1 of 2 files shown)


richfitz:
NOTE: to be really cron-able this needs to get authenticated with the vault without prompting

richfitz:
https://vimc.slack.com/archives/C835184N6/p1511886036000194

MartinEden:
What is the Slack message about? That just takes me to somewhere in #montagu-deploy...

richfitz:
That's an example message added in this PR (here: https://github.com/vimc/montagu/pull/54/files#diff-b52c3278761b1b251e47dccca3ca2b78R23)

MartinEden:
I think Slack was failing to take me to right message :(

richfitz:
Yeah, that's correct.  There's still an additional unresolved issue (that probably should be elsewhere) about how to get this running via cron.

Oh, one more thing:

4. sets up the test_user if the settings had required one

MartinEden:
That was the second thing I said :P

richfitz:
Ah sorry

PR #55, opened 2017-11-29
I1092 - bug with branches that fail issue lookup

littledice:
Fixes bug with branches that don't have tickets associated with them.

Simplified logic to treat branches as just one of:
1. a branch that *does* have a ticket associated with it, in which case we use that info in the release log
2. a branch that does *not* have a ticket associated with it, in which case we print a warning to the console,  and if the user continues list it in the 'other branches' section of the release log

i.e. don't bother distinguishing between the cases where a branch matches the regex but fails the issue lookup, and where a branch fails the regex

### DIFF
--- a/RELEASE_LOG.md
+++ b/RELEASE_LOG.md
@@ -140,3 +140,32 @@
 
 ## Other branches merged in this release
 
+# v0.8.5
+
+## Tickets
+* i1044_repository: Record set type when creating a new burden estimate set
+* i1067: Fix restore with annex
+* i1080: Find all instances of 'deterministic' and change to 'central' 
+* i1082: Add model_run_parameter_set to burden_estimate_set
+* i1083: Bug - run-development-apis.sh fails
+* i1084: Embed guidance within Montagu contribution portal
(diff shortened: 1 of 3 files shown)


RELEASE_LOG.md
```diff
@@ -137,3 +137,6 @@
 
 ## Tickets
 * i979: Push docker images to docker hub
+
+## Other branches merged in this release
+
```
littledice: just fixing our previously added manual edit to the release log, for consistency

scripts/release/make-release.py
```diff
@@ -44,15 +45,16 @@ def make_release_message(tag, branches_and_tickets):
         print("", file=msg)
         print("## Tickets", file=msg)
         for branch, ticket in branches_and_tickets:
-            if ticket:
+            if ticket == NOT_FOUND:
+                pass
+            else:
                 summary = ticket.get("summary")
```
littledice: the is the line that was throwing an error, if the ticket was 'NOT_FOUND'

PR #56, opened 2017-11-30
i1107 - simplify and document deployment from docker hub

richfitz:
As threatened this moves the top level deploy script from bash to python (because argument parsing in bash is not any fun).  The docker hub switch works by setting an environment variable and I've checked that makes it through to the subshell.  I've tried to reuse bits that are in scripts/releases (see the `sys.path.append("scripts/release")` line).


README.md
```diff
@@ -14,7 +14,22 @@ See [release process](ReleaseProcess.md)
 As root:
 
 1. `(cd src && pip3 install --user -r requirements.txt)`
-2. `./src/deploy.py`
+2. ./src/deploy.py
```
richfitz: This should go back to being in backticks!

PR #57, opened 2017-12-05
I1086 - Staging provisioning fixes

richfitz:
This started life as

https://vimc.myjetbrains.com/youtrack/issue/VIMC-1086

But I am doing a bunch of fixup work at the same time.  Sorry about the insane number of tiny commits but I was pushing things out to test on staging because I can't fit these machines on my workstation anymore.

* Adds a disk to each staging vm, like with the ci system and put docker's directories there.  This is to avoid the out-of-space errors that Martin reported in #montagu-deploy
* Instructions for rebuilding a machine, since I need to do that today.  This includes preserving and restoring the configuration, which sits outside of CI
* Set the hostname on each staging instance so you can tell where you are ssh'd into
* Customised `$PS1` for bash so it's _really_ easy to tell where you are
* Removes the self-signed certificate registry hack that we no longer need
* Clean up after vault provisioning
* Generalise the restore script to work for science and uat, with the nice benefit that they're now using the same system I plan on eventually testing periodically


staging/shared/restore
```diff
@@ -0,0 +1,69 @@
+#!/usr/bin/env bash
```
littledice: Think this script is missing a file extension?

staging/shared/record-montagu-configuration
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+set -x
```
littledice: the file also missing a file extension

staging/shared/restore
richfitz: I left it without because there was a script in there already without one and because this way we can swap them out for python scripts later without changing the documentation.  Editors _should_ detect the hashbang at the top

littledice: ok

PR #59, opened 2017-12-08
i1139 - support multiple montagu instances on a single host

richfitz:
The title of this PR is a bit of a lie; the vast majority of the PR could be phrased as

> Change `service.service` from being a global variable and initialise it with settings

This is a fairly large set of changes as a result and I can go back and split these two bits out if that seems necessary, though that will take a while of course.

I'm submitting this now before it gets any bigger but there are additional things that *could* be done (here or in a follow up PR/PRs)

* move registry configuration into the service object, either via the settings or a cli argument
* similar work to lift versions into the service object
* webapp_config.py is obsolete and can be removed I believe
* first time setup is obsolete - I'm raising Exception to prove this here, and can purge it too
* configuration file argument so that the second version can be deployed from the same (enormous) montagu checkout and kept in sync with the "main" deployment

One issue with testing this: we don't really have a full set of "deployment types" that is easy to this with.  I have locally tested:

* deploying a fake/minimal/vaultless instance
* the teamcity instance
* an instance much like science/uat (with both fake and readonly annexes)

Note that this introduces a lot of new variables - mostly port numbers.  The defaults will work for every previous instance I believe.

### DIFF
--- a/src/data_import.py
+++ b/src/data_import.py
@@ -3,39 +3,38 @@ from subprocess import check_output
 import backup
 import versions
 from docker_helpers import docker_cp
-from service import service
 from teamcity import save_artifact
 
 
-def do(settings):
-    source = settings["initial_data_source"]
+def do(service):
+    source = service.settings["initial_data_source"]
     print("Running initial data import with mode: {}".format(source))
     if source == "minimal":
         print("- Nothing to do (migrations will insert minimal data)")
     elif source == "test_data":
         local_path = get_artifact("montagu_api_generate_test_data", "test-data.sql", commit_hash=versions.api)
-        import_sql(local_path)
+        import_sql(db, local_path)
     elif source == "legacy":
         local_path = get_artifact("montagu_MontaguLegacyData_Build", "montagu.dump", "legacy-data.dump")
-        import_dump(local_path)
+        import_dump(db, local_path)
     elif source == "restore":
-        backup.restore(settings)
+        backup.restore(service)
     else:
         raise Exception("Unknown mode '{}'".format(source))
 
 
-def import_sql(sql_path):
+def import_sql(db, sql_path):
     print("- Copying {} to DB container and importing into database".format(sql_path))
     target_path = "/tmp/import.sql"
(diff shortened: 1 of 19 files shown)


MartinEden:
It would be great if the semantic changes and the refactor were separate - this is a big PR to review. I'll keep trying anyway, though, because I guess you don't have time to split it out given the deadline?

richfitz:
yeah, I'd really rather if we can avoid it

src/data_import.py
```diff
@@ -3,39 +3,38 @@
...
-    source = settings["initial_data_source"]
+def do(service):
+    source = service.settings["initial_data_source"]
     print("Running initial data import with mode: {}".format(source))
     if source == "minimal":
         print("- Nothing to do (migrations will insert minimal data)")
     elif source == "test_data":
         local_path = get_artifact("montagu_api_generate_test_data", "test-data.sql", commit_hash=versions.api)
-        import_sql(local_path)
+        import_sql(db, local_path)
```
MartinEden: Where does this `db` variable come from?

src/webapp_config.py
```diff
@@ -2,10 +2,9 @@
 
 from certificates import extract_certificates, cert_dir
 from docker_helpers import docker_cp
-from service import service
 
-
-def configure_webapps(keystore_password):
+## TODO: so far as I can see this is obsolete?
```
MartinEden: Yes, this is redundant - the webapps are behind nginx now, and that takes care of SSL.

src/data_import.py
richfitz: Nice catch - this is one of the untested bits.  That should be `service.db`

richfitz:
> I think we should add some unit tests for one or two bits of tricky logic

Absolutely.  The deploy always feels nasty to change because it's so easy to end up with breakage without tests there

richfitz:
Now that we aren't going down the route of running two instances on one machine, shall I remove the bits of this PR that support that and leave just the bits that eliminate the global `service` object?

richfitz:
Closing in favour of #63

PR #62, opened 2017-12-12
i1135 - add shiny server to the docker constellation

richfitz:
**Do not merge**

This is a starting point for discussion; at the least the port 3838 needs to be removed from the docker-compose.yml file as that would expose shiny apps to any user on the network without authentication.

To test, deploy with `"initial_data_source": "restore"` so that you have the actual reports then run

```
docker exec -it montagu_orderly_1 git checkout i1135
docker exec -it montagu_orderly_1 git pull

docker exec -it montagu_orderly_1 orderly run internal-2017-201710gavi-version2-import
docker exec -it montagu_orderly_1 orderly deploy-shiny /shiny
```

After that, go to http://localhost:3838/ (note **not** https) or http://localhost:3838/apps/ - there will be an app there!


richfitz:
The other relevant bits are in https://github.com/vimc/orderly/pull/4 and https://github.com/vimc/montagu-shiny/pull/1 with trivial changes to the dockerfiles for `montagu-orderly` and `orderly.server` still awaiting a PR (they point at the feature branch at the moment)

littledice:
It looks like the cert tool and proxy have updates but I don't see corresponding PRs for those? Have merged orderly and shiny, so you can update the references here to point back to master!

richfitz:
I'll double check, but don't merge this until all the auth bits are in (I think)

richfitz:
I've created PRs for the proxy and cert-tool now

littledice:
I've merged the other PRs so think we're good to point all submodules to master...

richfitz:
This is passing on teamcity now :)

PR #64, opened 2017-12-14
copy templates between volumes and mount volume into contrib portal

littledice:
Once this merged then merge: https://github.com/vimc/montagu-webapps/pull/127


src/setting_definitions/__init__.py
```diff
@@ -81,6 +81,13 @@ def vault_required(settings):
                               ("fake",       "Do not use passwords from the vault")
                           ]
                           ),
+    EnumSettingDefinition("db_annex_type",
+                          "How do we treat the annex database?",
+                          [
+                              ("fake", "Add a totally safe, but empty, version to the constellation"),
+                              ("readonly", "Read-only access to the real annex"),
+                              ("real", "Full access to the real annex: PRODUCTION ONLY")
+                          ]),
```
littledice: this moved to solve bug with this setting being referenced in `vault-required`, called below

richfitz: OK, that's also fixed in #63 but I'll deal with the conflict there is this goes in first

MartinEden:
Looks good to me!

src/service_config/contrib_portal_config.py
```diff
@@ -0,0 +1,14 @@
+from service import orderly_volume_name, template_volume_name
+from docker_helpers import copy_between_volumes
+
+
+def configure_contrib_portal(template_report_version):
+    print("Configuring contrib portal")
+    path_to_templates = "archive/internal-2017-burden-estimates-template/{}/*.csv".format(template_report_version)
```
richfitz: I'm not sure about this bit here:

* Perhaps we should do nothing if the template_report_version is the empty string - this must be needed on any instance without the orderly volume from backup?
* What perhaps make the format `<reportname>/<id>` so that this year's templates are not hard coded into the deploy
* We discussed multiple templates - does this report have to generate all types now?

littledice:
Changes made as requested @richfitz

teamcity-settings.json
```diff
@@ -12,5 +12,6 @@
     "add_test_user": false,
     "db_annex_type": "fake",
     "instance_name": "teamcity",
-    "notify_channel": ""
+    "notify_channel": "",
+    "template_reports":[]
```
richfitz: should this not be `"template_reports": ""` (empty string rather than an empty array?)

littledice: No because these are the parsed settings - if the user passes in a empty string on the command line, it will get saved here as an empty array

richfitz: oh of course!

richfitz:
I'm just thinking about this a bit more, and I think we might have discussed this, but I'm wondering if it might make more sense to tie the report versions to the release and the config option could be whether or not to deploy them.

So have a file in the repo which has a newline separated set of reports that are the "current" reports.  And then replace `template_reports` with a boolean option `include_template_reports` perhaps.

This seems more similar to our current approach and increases the chance that we'll get the same content deployed across different instances

littledice:
I'm testing this locally (taking a while as restoring data first) but does this look right to you?

src/deploy.py
```diff
@@ -105,7 +105,9 @@ def configure_montagu(service, is_first_time):
                   service.settings["hostname"], send_emails)
     configure_reporting_api(service, token_keypair_paths)
     configure_proxy(service, cert_paths)
-    configure_contrib_portal(service, service.settings["template_reports"])
+
+    if service.settings["inclue_template_reports"]:
```
richfitz: typo here (`include_template_reports` not `inclue_template_reports`)

src/service_config/contrib_portal_config.py
```diff
@@ -14,3 +15,10 @@ def add_templates_to_contrib_portal(service, path_to_templates):
     print("- Copying burden estimate templates from orderly to contrib portal")
     copy_between_volumes(service.volume_name("orderly"), service.volume_name("templates"), path_to_templates)
 
+
+def get_template_report_versions():
+    with open("template_report_versions") as f:
+        value = f.read()
+    if len(value) == 0:
+        return []
+    return value.split()
```
richfitz: Consider 

```python
[x.strip() for x in value.split("\n") if x]
```

to go with newline delimited and to strip the trailing newline (and over any blank lines)

richfitz: That also means you can avoid the early exit

src/setting_definitions/array.py
```diff
@@ -0,0 +1,12 @@
+from .definition import SettingDefinition
+
+
+class ArraySettingDefinition(SettingDefinition):
```
richfitz: And this can come out now, right? (along with the corresponding import).  Otherwise this looks cool - it's a really small set of changes for something that is going to make the images way nicer to deal with

PR #65, opened 2017-12-14
i1205 - include static website into deployment

richfitz:
This builds on the work in https://github.com/vimc/vimc-website/pull/1 and https://github.com/vimc/montagu-proxy/pull/10 - relatively little is needed at the montagu end.  This adds one additional docker volume to the constellation (but no more containers).  There's a small script that will build the site - it can be run without deploying and is run at the end of deploy.  Builds after the first will be much faster, so with persistent volumes this will not slow things down more than about 1s

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -44,6 +44,8 @@ services:
       - admin
       - contrib
       - report
+    volumes:
+      - vimc_website_volume:/vimc-website
     command: ${MONTAGU_PORT} ${MONTAGU_HOSTNAME}
   orderly:
     image: ${MONTAGU_REGISTRY}/montagu-orderly:${MONTAGU_ORDERLY_VERSION}
@@ -54,4 +56,5 @@ services:
 volumes:
   db_volume:
   orderly_volume:
+  vimc_website_volume:
   emails:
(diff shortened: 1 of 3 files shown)


richfitz:
After running this, then go to http://localhost:8080/ and the main website will be shown

docker-compose.yml
```diff
@@ -37,13 +37,16 @@ services:
     image: ${MONTAGU_REGISTRY}/montagu-reverse-proxy:${MONTAGU_PROXY_VERSION}
     ports:
       - "${MONTAGU_PORT}:${MONTAGU_PORT}"
-      - 80:80
+      - "80:80"
+      - "8080:8080"
```
littledice: I don't think there's any point mapping 8080 here... see proxy PR https://github.com/vimc/montagu-proxy/pull/10#pullrequestreview-83774227

richfitz: I'm a bit confused about how to proceed here - I've exposed this on 8080 just to check that things work.  Once we can restructure the nginx config I agree that this should come out but I'm not sure how to actually do that!

littledice: Ok yeah I'm with you, maybe you're right that this is the first step.

PR #67, opened 2017-12-20
fetch before getting latest tag

littledice:


scripts/release/helpers.py
```diff
@@ -7,3 +7,8 @@ def run(cmd, working_dir=None):
     result = subprocess.run(parts, check=True, stdout=subprocess.PIPE,
                             universal_newlines=True, cwd=working_dir)
     return result.stdout.strip()
+
+
+def fetch():
+    print("Fetching from remote...")
+    run("git fetch --tags --all")
```
richfitz: Why `--all` here? We don't typically use more than one remote?

littledice: This function has just moved but not been altered, I don't know why it was originally written like this!

richfitz: ah, of course :)

PR #71, opened 2018-01-10
I1126 add annex credentials to api

littledice:
One thing I wasn't sure of was whether we want to add an api specific user to the annex? I've just passed through the root user credentials...

### DIFF
--- a/src/deploy.py
+++ b/src/deploy.py
@@ -105,7 +105,7 @@ def configure_montagu(service, is_first_time):
     send_emails = service.settings["password_group"] == 'production'
     add_annex = service.settings["db_annex_type"] != 'readonly'
     configure_api(service, passwords['api'], token_keypair_paths,
-                  service.settings["hostname"], send_emails, add_annex, annex_root_user.name, annex_root_user.password)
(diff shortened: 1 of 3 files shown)


src/deploy.py
```diff
@@ -94,15 +94,18 @@ def configure_montagu(service, is_first_time):
...
 
     # Push secrets into containers
     cert_paths = get_ssl_certificate(service.settings["certificate"])
     token_keypair_paths = get_token_keypair()
 
     send_emails = service.settings["password_group"] == 'production'
+    add_annex = service.settings["db_annex_type"] != 'readonly'
     configure_api(service, passwords['api'], token_keypair_paths,
-                  service.settings["hostname"], send_emails)
+                  service.settings["hostname"], send_emails, add_annex, annex_root_user.name, annex_root_user.password)
```
MartinEden: Why not just pass the user object through as a single argument, instead of splitting it up into separate username and password args?

MartinEden:
Also merge conflicts.

MartinEden:
Still needs conflicts resolving.

PR #72, opened 2018-01-12
i1315 - set up ssh on staging vms

richfitz:
This is needed for pulling montagu-data and montagu-reports

### DIFF
--- a/scripts/setup_ssh_keys.sh
+++ b/scripts/setup_ssh_keys.sh
@@ -1,0 +1,20 @@
+#!/usr/bin/env bash
+set -e
+
+export VAULT_ADDR='https://support.montagu.dide.ic.ac.uk:8200'
+if [ ! -f ~/.vault-token ]; then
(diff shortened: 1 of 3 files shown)


richfitz:
Mostly used here: https://github.com/vimc/montagu/tree/master/staging#to-rebuild-a-vm

richfitz:
These scripts are used only when setting up a VM.  The idea is that we access the vault on startup and then all the bits that are needed for use of montagu are set up.  This is:

- a copy of the montagu repo, ready for deployment
- a copy of montagu-orderly, to run reports
- a copy of montagu-data for testing data imports (primarily useful on science)

The last requires ssh keys because it is a private repo.  There are private repos in the vault, which this script accesses and then sets up on the machine so that the clone works.

This is tested programmatically in `restore-test` (not yet run nightly, but once the backup work is done that was my intention).

This was written mostly because everytime we rebuild the machines it's a faff to download the private repos and I wanted a script to do so.

Let me know if that's not clear enough and we can discuss further in person perhaps

PR #78, opened 2018-01-29
i1366: Restart Montagu on reboot 

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-1366

First review and merge API change

### DIFF
--- a/src/test.py
+++ b/src/test.py
@@ -53,6 +53,9 @@ def webapp_integration_tests():
 
 if __name__ == "__main__":
     if len(sys.argv) > 1 and sys.argv[1] == "--run-tests":
+        # Imitate a reboot of the system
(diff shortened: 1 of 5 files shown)


src/test.py
```diff
@@ -53,6 +53,9 @@ def work():
 
 if __name__ == "__main__":
     if len(sys.argv) > 1 and sys.argv[1] == "--run-tests":
+        # Imitate a reboot of the system
+        print("Restarting Docker", flush=True)
```
richfitz: Can we put this behind an extra flag (that might drag in docopt) because restarting docker is pretty antisocial :)

MartinEden: Good idea.

PR #79, opened 2018-02-02
i1412: Update diagram to include shiny and caddy

richfitz:


MartinEden:
Seems reasonable. Maybe move the "backup daemon" and "database clients" boxes to the left so as to straighten the lines up a bit? But thanks for updating this!

richfitz:
This is updated nows

PR #80, opened 2018-02-07
i1472: Add copies of settings

richfitz:
This is a starting point for discussion.  I want to store the settings for production/science/uat in these files like we do for teamcity because otherwise it's annoying to remember everything when rebuilding a machine

What is not in here is anything to indicate how we should keep these up to date.  Options:

* compare the settings against the appropriate file and warn as they diverge
* ./deploy.py takes a settings file argument
* other?


littledice:
I like the idea of changing settings always being a case of committing to the repo. But I'm not sure about taking a settings file as an argument, unless we could also match it against the hostname somehow, to avoid deploying the wrong thing? Just thinking about how to minimise user error.

Maybe we could have an independent `copy-config` script that copies the config you want into the right place? So changing the config would be a case of committing to this repo, then running that `copy-config` script once? Then you'd have plenty of opportunity to check that you're using the right config before deploying, and you wouldn't have to do it often.

richfitz:
I've redone this with the settings from the current machines, and with `restore-test.json` reflecting science for now

richfitz:
This is work that would be nice to do here, but I think can wait: https://vimc.myjetbrains.com/youtrack/issue/VIMC-2277

PR #81, opened 2018-02-07
I1434 - copy guidance reports into contrib container

littledice:
Have confirmed this works locally, i.e. the report paths are valid and result in exposing:

`../contribution/guidance/child-mortality.html`
`../contribution/guidance/over80.html`
`../contribution/guidance/kosovo_demography.html`
`../contribution/guidance/marshall_demography.html`
`../contribution/guidance/tuvalu_demography.html`

One question I have here is whether the info in these reports is at all sensitive?
With the templates we don't worry about them being publicly viewable because they contain no data; I don't know if that is true here or whether we'd want to lock them down to logged in users (I guess we'd have to use Caddy if so!!)
@weshinsley do you know the answer?

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -26,6 +26,7 @@ services:
       - api
     volumes:
       - template_volume:/usr/share/nginx/html/templates
+      - guidance_volume:/usr/share/nginx/html/guidance
   admin:
     image: ${MONTAGU_REGISTRY}/montagu-admin-portal:${MONTAGU_ADMIN_PORTAL_VERSION}
     depends_on:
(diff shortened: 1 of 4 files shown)


littledice:
@MartinEden pointed out that there is an alternative to this approach, though it'd be more work to implement - we can serve the files via the reporting API;  would need to add a new permission, like `guidance.read` and assign to all modellers. But if we do need to lock these files down, maybe we ditch this and go with serving via the reporting API.

richfitz:
So with that approach we'd avoid copying any files over?  That seems like a definite plus

littledice:
Yeah. This is kind of the laziest implementation possible, but ultimately serving via the API will be better.

littledice:
I've had the OK on this from @tinigarske and @weshinsley - the reports in question are all re demography and therefore not sensitive. So lets do this simple fix for now.

PR #82, opened 2018-02-08
i1466: logging via systemd

richfitz:
Please note - this requires docker-compose 1.17.0 or greater.  If you do not have that version you will get an error like:

```
The Compose file './../docker-compose.yml' is invalid because:
Additional properties are not allowed ('x-logging' was unexpected)

You might be seeing this error because you're using the wrong Compose file version. Either specify a supported version ("2.0", "2.1", "3.0", "3.1", "3.2") and place your service definitions under the `services` key, or omit the `version` key and place your service definitions at the root of the file to use version 1.
For more on the Compose file format versions, see https://docs.docker.com/compose/compose-file/
```

I have not added logging to the (fake) annex because we don't use that in production.  I can add it if needed but it will require duplicating the contents of the `x-logging` section into the annex configuration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # Montagu
 ## Prerequisites
 * [Docker Community Edition](https://docs.docker.com/engine/installation/) 
-* [Docker Compose](https://docs.docker.com/compose/install/)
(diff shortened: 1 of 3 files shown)


littledice:
We'll need to update `docker-compose` on prod - I'll do so now

PR #87, opened 2018-03-19
I1559 - enable streaming replication from database

richfitz:
This is about half of the deploy work that needs doing.  It does not do anything about a restore but that is likely to be quite a bit easier because it will be largely removing code perhaps?  But I'd like us to see how doing the streaming works in practice.

To work it will need the updated version of montagu-db too and that must be merged first (https://github.com/vimc/montagu-db/pull/66: there is some small configuration changes there).

This PR is complicated by interacting with the user management.  The replication needs two more users and these need creating with special options (one is a superuser the other is a replication user).  The other complication is that the current version of the python docker package does not propagate exit code failure so I've had to do that with a wrapper (see the comment there for more details).  

This PR adds one additional configuration option but I would also be happy to set up streaming replication unconditionally.  It should not be problematic really I think, and that might simplify the user management a little?


MartinEden:
@richfitz Please fix merge conflicts

richfitz:
The merge conflict is the db submodule - to resolve this I need to wait for the montagu-db PR to become mergeable and then I'll merge master in here

PR #93, opened 2018-04-20
i1682: Install and schedule bb8 during deploy 

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-1682

This also adds the ability to restore from bb8 as the initial data source. But I've only got bb8 working for the orderly volume, which is the easy bit. There are other tickets for database backup via barman, and database restore from bb8.

First merge https://github.com/vimc/montagu/pull/92 and https://github.com/vimc/bb8/pull/20, and update the montagu-bb8 submodule.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -31,3 +31,6 @@
 [submodule "submodules/shiny"]
 	path = submodules/shiny
 	url = https://github.com/vimc/montagu-shiny
+[submodule "montagu-bb8"]
+	path = montagu-bb8
+	url = https://github.com/vimc/montagu-bb8
--- a/src/bb8_backup.py
+++ b/src/bb8_backup.py
@@ -32,7 +32,7 @@ def backup():
 @requires_bb8_setup
 def schedule():
     print("Scheduling bb8 backup")
-    run("./schedule.sh", cwd=bb8_dir, check=True)
+    run("./schedule", cwd=bb8_dir, check=True)
 
 
 @requires_bb8_setup
--- a/src/bb8_backup.py
+++ b/src/bb8_backup.py
@@ -11,13 +11,15 @@ def requires_bb8_setup(func):
         if not finished_setup:
             setup()
         return func(*args, **kwargs)
+
     return decorated
 
 
 def setup():
     global finished_setup
-    print("- Configuring and installing bb8 backup service with these "
-          f"targets: {targets}")
+    template = "- Configuring and installing bb8 backup service with these " \
+               "targets: {}"
+    print(template.format(targets))
     args = ["./setup.sh", "../config.json"] + targets
     run(args, cwd=bb8_dir, check=True)
     finished_setup = True
--- a/src/bb8_backup.py
+++ b/src/bb8_backup.py
@@ -1,0 +1,41 @@
+from subprocess import run
+
+targets = ["orderly"]
+finished_setup = False
+bb8_dir = "../montagu-bb8/bb8"
+
+
+def requires_bb8_setup(func):
+    def decorated(*args, **kwargs):
+        global finished_setup
+        if not finished_setup:
+            setup()
+        return func(*args, **kwargs)
+    return decorated
+
+
+def setup():
+    global finished_setup
+    print("- Configuring and installing bb8 backup service with these "
+          f"targets: {targets}")
+    args = ["./setup.sh", "../config.json"] + targets
+    run(args, cwd=bb8_dir, check=True)
+    finished_setup = True
+
+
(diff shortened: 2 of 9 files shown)


.gitmodules
```diff
@@ -31,3 +31,6 @@
 [submodule "submodules/shiny"]
 	path = submodules/shiny
 	url = https://github.com/vimc/montagu-shiny
+[submodule "montagu-bb8"]
```
richfitz: Is there a reason, aside from `backup` being at the top level, not to put this under `submodules`?

src/bb8_backup.py
```diff
@@ -0,0 +1,43 @@
...
+@requires_bb8_setup
+def backup():
+    print("Performing bb8 backup")
+    run(["bb8", "backup"], check=True)
+
+
+@requires_bb8_setup
+def schedule():
+    print("Scheduling bb8 backup")
+    run("./schedule", cwd=bb8_dir, check=True)
```
richfitz: Does this mean we still need to deploy as root? What does the workflow look like if this step fails?

For barman, I allowed this part to fail and printed a message to run.  One way of managing that would be to add a flag to `schedule` that controls if it must work or not and printing the instructions there

src/bb8_backup.py
```diff
@@ -0,0 +1,43 @@
...
+
+    return decorated
+
+
+def setup():
+    global finished_setup
+    template = "- Configuring and installing bb8 backup service with these " \
+               "targets: {}"
+    print(template.format(targets))
+    args = ["./setup.sh", "../config.json"] + targets
```
richfitz: this path algebra here is confusing me - this is `../config.json` from the point of view of `../montagu-bb8/bb8` which ends up being `../montagu-bb8/config.json` I think.  Would it be easier (and allow more easy moving of modules at a later date if wanted) to run the path of the config through `os.path.abspath` and pass that through as the argument?

src/bb8_backup.py
```diff
@@ -0,0 +1,43 @@
...
+    run(args, cwd=bb8_dir, check=True)
+    finished_setup = True
+
+
+@requires_bb8_setup
+def backup():
+    print("Performing bb8 backup")
+    run(["bb8", "backup"], check=True)
```
richfitz: Out of scope for this ticket but if we provided bb8 as a python module we'd be able to centralise a lot of the logic in this file for use on subsequent projects?  If that sounds useful I can open a ticket

src/bb8_backup.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+@requires_bb8_setup
+def schedule():
+    print("Scheduling bb8 backup")
+    run("./schedule", cwd=bb8_dir, check=True)
+
+
+@requires_bb8_setup
+def restore():
```
richfitz: This no longer needs `service` (vs the duplicati version) which means that the volume names are duplicated between the docker-compose.yml and the montagu-bb8 configuration https://github.com/vimc/montagu-bb8/blob/master/config.json

Can we (at some point, not necessarily here) move to generating the json used by bb8 from the configuration as used by montagu?

.gitmodules
MartinEden: I feel like the things in `submodules` are the things that we just include as submodules so we can do git magic and track which docker images to deploy. Whereas `backup` and `bb8` are different in that we are actually using the contents of the submodules to do things. So I wanted to keep them separate. I agree that a folder named `submodules` is not exactly the clearest way to do that, but moving submodules around is a pig.

src/bb8_backup.py
MartinEden: Yeah, currently, if there is anything that bb8 needs to do as root (which there will only sometimes be) I intend the deploy to fail and you to rerun it as root. I fear that a warning message will just get lost?

src/bb8_backup.py
MartinEden: Like this?

src/bb8_backup.py
MartinEden: That sounds good :)

src/bb8_backup.py
MartinEden: That sounds like a good idea: https://vimc.myjetbrains.com/youtrack/issue/VIMC-1693

.gitmodules
richfitz: yeah, that makes sense

src/bb8_backup.py
richfitz: https://vimc.myjetbrains.com/youtrack/issue/VIMC-1699

PR #98, opened 2018-04-27
i1724: Set root password unconditionally

richfitz:
As used on uat

We should eventually move to using real passwords (can be done outside of PR with a configuration change) and then eventually to refusing to restore with fake passwords

### DIFF
--- a/src/database.py
+++ b/src/database.py
@@ -277,9 +277,9 @@ def setup(service, annex_settings):
     print("- Scrambling root password")
     if password_group is not None:
         root_password = GeneratePassword().get()
-        set_root_password(service, root_password)
     else:
         root_password = 'changeme'
+    set_root_password(service, root_password)
 
     print("- Getting user configurations")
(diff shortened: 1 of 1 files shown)


src/database.py
```diff
@@ -277,9 +277,9 @@ def setup(service, annex_settings):
     print("- Scrambling root password")
     if password_group is not None:
         root_password = GeneratePassword().get()
-        set_root_password(service, root_password)
     else:
         root_password = 'changeme'
+    set_root_password(service, root_password)
```
littledice: Do we actually need to have the root user available to us on UAT? What do we use it for atm? Maybe we should just scramble it on UAT also?

richfitz: good question! we could just all of this unconditionally. Any thoughts @MartinEden?

littledice: Oh I got overexcited and merged this. We could do this as part of VIMC-1726

PR #103, opened 2018-06-14
I1848 - hardening of SSL

littledice:
We now have another private key stored in the vault, a DHE parameter, as recommend by this blog https://raymii.org/s/tutorials/Strong_SSL_Security_On_nginx.html#Forward_Secrecy_&_Diffie_Hellman_Ephemeral_Parameters

I've generated 3, 2 of them are stored in the vault under `ssl/v2/production/dhparam` and `ssl/v2/support/dhparam`, and the third included here for the self-signed case.


littledice:
You're totally right, it should be. Added to the proxy README: https://github.com/vimc/montagu-proxy/blob/5835cd6cfaa490e6c85a62d5ec0ab95965a5dbb4/README.md

PR #104, opened 2018-06-20
i1882: option to update volumes on deploy

richfitz:
Requires https://github.com/vimc/montagu-db/pull/82

We did discuss logic and configuration but I don't remember the outcome!  This seems like the least awkward way to me: a new boolean, which is used only when

* we are already using bb8 as the data source
* we are not backing up to bb8
* it's not a first time deployment

On my desktop with slow disks it is taking up to a minute or two for the db to come up after recovery is started, so I have added a call to montagu-wait into the deploy step.  I've waited for (up to) an hour but we can update if needed.  If the db does not come up at all (e.g., container fails to start or dies) this will fail much faster; the wait should only be during postgres' maintenance.


src/deploy.py
```diff
@@ -71,7 +71,9 @@ def _deploy():
     # BB8 restore
     data_exists = (not is_first_time) and service.settings["persist_data"]
     if settings["initial_data_source"] == "bb8_restore":
-        if data_exists:
+        data_update = service.settings["update_on_deploy"] and \
+                      not service.settings["bb8_backup"]
+        if data_exists and not data_update:
             print("Skipping bb8 restore: 'persist_data' is set, "
                   "and this is not a first-time deployment")
```
MartinEden: This message should be updated. Ideally it should tell which half of the `and` predicate failed, so the user has a clear idea of what they need to change if this behaviour was not expected.

e.g. "Skipping bb8 restore: bb8_backup is set to True"

MartinEden: Actually, this logic is quite complicated. Maybe we are conflating two things here. Maybe we should remove bb8 from the `initial_data_source` options, as I think it's not needed if we have this `update_on_deploy` mechanism. bb8 restore already happens at a different point in the process.

littledice: Agree the message should be updated, but the logic seems simple enough and correct to me - if it's a first time deploy, or if `update_on_deploy` is true (and this isn't the source of truth for bb8, i.e. production), then run `bb8 restore`.

MartinEden: But is there any need for "initial_data_source" to be bb8? Doesn't the `update_on_deploy` mechanism supersede that?

richfitz: What about for a first time deployment?  This should differ for a new support and production instance

MartinEden: Okay fair enough. So if we're doing a clean install of production then we need initial_data_source to be bb8, but we don't want then to update every deploy. Whereas on support/science we want update_on_deploy.

PR #106, opened 2018-06-22
i1849: New SSL certificate

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-1849


certs/production/montagu.vaccineimpact.org.crt
```diff
@@ -1,110 +1,39 @@
 -----BEGIN CERTIFICATE-----
-MIIFVzCCBD+gAwIBAgIJAN/wllrH5ZvyMA0GCSqGSIb3DQEBCwUAMIG0MQswCQYD
```
littledice: Why did this have 3 certificates before, and now just 1?

MartinEden: Very good catch. It turns out the thing I was using to inspect the certificate file was only printing out the first certificate in the bundle, leading me to think it wasn't a bundle. I've improved that now, and have (I think) properly created a new bundle.

MartinEden:
We don't need a new certificate for support - that one's internal, provided by Imperial (not GoDaddy) and lasts until 2020.

MartinEden:
Here is the diff between the old certificate bundle and new certificate bundle in human-readable form:
[diff.txt](https://github.com/vimc/montagu/files/2132640/diff.txt)

PR #111, opened 2018-08-23
i2121: Auto update

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-2121


scripts/auto-update-and-commit.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+set -ex
+./src/update_versions_to_latest_master.py
+# https://stackoverflow.com/a/3879077/777939
+git diff-index --quiet HEAD -- && echo "No changes detected" && exit 0;
+echo "Changes detected"
+
+git remote set-url origin git@github.com:vimc/montagu
+git push --delete origin latest || true
+git branch -d latest || true
```
richfitz: Can you make this `-D` to `-d` to avoid any issue with branch deletion? Or does deleting upstream make this always safe? Git will refuse to delete a branch that is ahead of master so I think this will fail when it is run on a machine for the *second* time that changes in submodules are detected

scripts/auto-update-and-commit.sh
```diff
@@ -0,0 +1,15 @@
...
+git checkout -b latest
+
+git add submodules
+git commit -m "Auto: Update versions to latest"
+git push -u origin latest
```
richfitz: So as I understand it we have the ordinary deploy task running on all branches, but this one just running `master` right? So on the main deploy task we should trigger a staging machine after successful deploy iff the branch is `latest`. Do we already have a separate ticket for that?

scripts/auto-update-and-commit.sh
MartinEden: The help says:

```
-d, --delete
           Delete a branch. The branch must be fully merged in its upstream
           branch, or in HEAD if no upstream was set with --track or
           --set-upstream-to.
```

I would guess that if the upstream branch has been deleted this will work. Maybe we try it out (which is tricky to do immediately, as we need more changes to be made in the submodules) and then change if necessary?

scripts/auto-update-and-commit.sh
MartinEden: I've raised https://vimc.myjetbrains.com/youtrack/issue/VIMC-2123 for this.

scripts/auto-update-and-commit.sh
littledice: I'm pretty sure `-d` doesn't work when the local branch is ahead of master, regardless of upstream.

richfitz: Sorry my comment is a mess! Can you change from `-d` to `-D`.

```
       -D
           Shortcut for --delete --force.
       -f, --force
           [...] In combination with -d (or --delete), allow deleting the
           branch irrespective of its merged status.
```

richfitz: So it's the merged status that matters in addition to upstream. But in our case we just want the branch gone and `-D` will do the same as `-d`. Or we can leave it as `-d` and see if it actually matters in practice

MartinEden: I've changed it to `-D`

MartinEden:
So, happy to merge this, @richfitz?

richfitz:
Yup, thanks!

PR #118, opened 2018-10-03
i2219: Enable production db config

MartinEden:
https://vimc.myjetbrains.com/youtrack/issue/VIMC-2219

I also:

* Made `montagu-deploy.json` more readable, by sorting indenting and sorting the keys
* Added defaults to a bunch of the settings, to make it easier to quickly set up a local test instance from scratch (which is when I am most often repeatedly answering these questions)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -38,7 +38,7 @@ services:
     image: ${MONTAGU_REGISTRY}/montagu-db:${MONTAGU_DB_VERSION}
     restart: always
     logging: *log-journald
-    command: ${MONTAGU_DB_CONF:-/etc/montagu/postgresql.conf}
+    command: ${MONTAGU_DB_CONF}
     ports:
       - "5432:5432"
     volumes:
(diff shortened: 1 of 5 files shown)


docker-compose.yml
```diff
@@ -38,6 +38,7 @@ services:
     image: ${MONTAGU_REGISTRY}/montagu-db:${MONTAGU_DB_VERSION}
     restart: always
     logging: *log-journald
+    command: ${MONTAGU_DB_CONF:-/etc/montagu/postgresql.conf}
```
richfitz: Is there a reason not to _always_ provide the configuration here? So for non-production environments pass through `/etc/montagu/postgresql.conf` in `compose.py` (around line 57/58)

MartinEden:
I'm totally in favour of https://github.com/vimc/montagu/pull/80

docker-compose.yml
MartinEden: You're right. I had a vague idea about adding defaults to the compose file so we could use it outside of the deploy tool. But that's clearly not practical without changing all the other env vars, and quite possibly not desirable either.

PR #119, opened 2018-10-04
i2116: Update barman instructions

richfitz:
Probably best to hold off merging this until after confirming it works!


richfitz:
see https://github.com/vimc/montagu-db/pull/101

PR #121, opened 2018-10-09
I2160 - Auto update script

weshinsley:
Running this script on linux hardware will enable automatic installation of security patches, and requests for reboots will be posted to the montagu-deploy channel via the autoupdate-webhook.

Installed and running on annex, support and production, which are all currently running 16.04.5 LTS


scripts/enable_automatic_updates.sh
```diff
@@ -0,0 +1,56 @@
...
+echo '
+APT::Periodic::Update-Package-Lists "1";
+APT::Periodic::Unattended-Upgrade "1";
+' >/etc/apt/apt.conf.d/20auto-upgrades
+
+sed -i.bak 's|//Unattended-Upgrade::Remove-Unused-Dependencies "false"|Unattended-Upgrade::Remove-Unused-Dependencies "true"|' /etc/apt/apt.conf.d/50unattended-upgrades
+
+sed -i.bak 's|//Unattended-Upgrade::Automatic-Reboot "false"|Unattended-Upgrade::Automatic-Reboot "false"|' /etc/apt/apt.conf.d/50unattended-upgrades
+
+rm /etc/apt/apt.conf.d/50unattended-upgrades.bak
```
MartinEden: If you're running this script (as opposed to manually working through it line by line) is there any point in using `.bak` in the `sed` command and then immediately removing the .bak file?

scripts/enable_automatic_updates.sh
```diff
@@ -0,0 +1,56 @@
...
+apt-get install unattended-upgrades curl
+
+echo '
+APT::Periodic::Update-Package-Lists "1";
+APT::Periodic::Unattended-Upgrade "1";
+' >/etc/apt/apt.conf.d/20auto-upgrades
+
+sed -i.bak 's|//Unattended-Upgrade::Remove-Unused-Dependencies "false"|Unattended-Upgrade::Remove-Unused-Dependencies "true"|' /etc/apt/apt.conf.d/50unattended-upgrades
+
+sed -i.bak 's|//Unattended-Upgrade::Automatic-Reboot "false"|Unattended-Upgrade::Automatic-Reboot "false"|' /etc/apt/apt.conf.d/50unattended-upgrades
```
MartinEden: This line doesn't do anything, right? I assume you've put it here to make it clear how to enable automatic reboot in the future. It might be clearer to replace it with a line that actually does enable automatic reboot, and then comment it out?

scripts/enable_automatic_updates.sh
```diff
@@ -0,0 +1,56 @@
...
+rm /etc/apt/apt.conf.d/50unattended-upgrades.bak
+
+echo '
+APT::Periodic::Update-Package-Lists "1";
+APT::Periodic::Download-Upgradeable-Packages "1";
+APT::Periodic::AutocleanInterval "1";
+APT::Periodic::Unattended-Upgrade "1";
+APT::Periodic::Verbose 2;
+APT::Periodic::RandomSleep 1;
+'> /etc/apt/apt.conf.d/10periodic
```
MartinEden: Is there a good place to link to docs for what each value does? They're pretty self explanatory, but it might still be good.

scripts/enable_automatic_updates.sh
```diff
@@ -0,0 +1,56 @@
...
+if [ -f /var/run/reboot-required ]; then
+
+  hostname=$(uname -n)
+  if [ $hostname = "fi--didevimc01" ]; then
+    hostname="production"
+  elif [ $hostname = "fi--didevimc02" ]; then
+    hostname="support"
+  elif [ $hostname = "wpia-didess1" ]; then
+    hostname="annex"
+  fi
```
MartinEden: Really nice to have the human friendly names, but also a fallback :) :+1:

richfitz:
Re: `montagu-machine` - let's do that as a separate refactor given this is hear already

scripts/enable_automatic_updates.sh
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env bash
+
+apt-get install unattended-upgrades curl
```
richfitz: add `apt-get update` first, then add `-y` argument to install

scripts/enable_automatic_updates.sh
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env bash
+
```
richfitz: Consider `set -e`

scripts/enable_automatic_updates.sh
```diff
@@ -0,0 +1,56 @@
...
+if [ "$VAULT_AUTH_GITHUB_TOKEN" = "" ]; then
+    echo -n "Please provide your GitHub personal access token for the vault: "
+    read -s token
+    echo ""
+    export VAULT_AUTH_GITHUB_TOKEN=${token}
+fi
+vault login -method=github
+SLACK_WEBHOOK=$(vault read -field=password secret/slack/autoupdate-webhook)
+
+echo '#!/bin/sh
```
richfitz: I feel this would be clearer and easier to edit with a heredoc string (e.g.

```
echo <<< EOF > /usr/local/sbin/check_reboot_required.sh
(all the contents here)
EOF
```

(same with the other big one above)

scripts/enable_automatic_updates.sh
MartinEden: I see. Maybe add a comment to these two lines, as it's quite easy to miss? Or maybe I'm just not paying enough attention.

scripts/enable_automatic_updates.sh
weshinsley: Not sure how you get this to work - my bash script is trying to execute (all the contents here)

weshinsley: Ah - << instead of <<< ?

weshinsley: No... that creates an empty file...

weshinsley: Closest I can get is 
```
tee /etc/apt/apt.conf.d/10periodic <<EOF > /dev/null
(all the contents here)
EOF
```
And I can then put the quotes in literally, but have to back-slash the $

scripts/enable_automatic_updates.sh
weshinsley: Ok!

scripts/enable_automatic_updates.sh
weshinsley: Added

scripts/enable_automatic_updates.sh
weshinsley: I've pre-fixed a comment linking to the ubuntu docs page.

scripts/enable_automatic_updates.sh
weshinsley: True - .bak no longer needed!

scripts/enable_automatic_updates.sh
weshinsley: Comment added - yes, easy to miss, the "//" or assume it's a syntactic part of the s|//

scripts/enable_automatic_updates.sh
weshinsley: (back-slash $ when you want it to stay a $ I mean. The final $SLACK_WEBHOOK isn't back-slashed, because we want to echo the value into the script)

richfitz: sorry wes, the heredoc should be

```
cat << EOF > target
text here
EOF
```

PR #125, opened 2018-12-06
add release to UAT to ReleaseProcess.md

littledice:


ReleaseProcess.md
```diff
@@ -9,7 +9,7 @@
    This script will warn you if any tickets are merged in that are not "Ready 
    to Deploy". It will automatically push the tag and commit.
 3. Check build has passed in Teamcity
-4. [Connect to the UAT machine and deploy there](staging/README.md)
```
littledice: this actually just contains info about provisioning a staging machine, not about release

ReleaseProcess.md
```diff
@@ -18,3 +18,24 @@
    1. `cd /montagu/deploy`
    1. `./deploy.py --docker-hub`
 7. Use `RELEASE_LOG.md` to know which tickets to update to the 'Deployed' status
+
+## Deploying to UAT
+
+1. `ssh support.montagu`
```
richfitz: I usually combine step 1 & 2 with `ssh vagrant@support.montagu`, which I think is a touch cleaner (`sudo` always being ever so slightly weird) - is that worth mentioning as an approach here?

ReleaseProcess.md
```diff
@@ -18,3 +18,24 @@
...
+2. Become the vagrant user and cd into their home directory `sudo su vagrant && cd`
+3. Run `./uat.sh` which will give you a shell inside the virtual machine.
+4. `cd montagu`
+5.  2 options here:
+
+    i) To deploy the latest tagged release, run `sudo -E ./deploy.py`
+
+    ii) To run a specific branch, for example to run the latest master branch:
+    ```
+    sudo git checkout master && git pull && git submodule update --recursive
```
richfitz: I'm sure we're all git-savvy enough on the team to work it out, but for a feature branch that is not yet present on the machine this will not work. Can I suggest:

```
git fetch && git checkout <branchname> && git merge && git submodule update --recursive
```

and we _really_ need to stop the `sudo` madness here! (see [VIMC-1747](https://vimc.myjetbrains.com/youtrack/issue/VIMC-1747) - I might look at this on the next automation block of work)

richfitz: Perhaps this is actually something worth baking into the deploy tool - logged as [VIMC-2430](https://vimc.myjetbrains.com/youtrack/issue/VIMC-2430)

littledice: I know, it's not nice, and good point re branch, will update

PR #138, opened 2019-01-16
i2253: new script to create our latest branch

richfitz:


richfitz:
can only really be tested by deploying eventually and seeing changes in the various trigger modes:
* everything up to date - no build
* latest has been built and nothing else has changed - no build
* submodules updated - build
* montagu master itself is updated - build

scripts/create-latest
```diff
@@ -0,0 +1,115 @@
...
+
+
+def subprocess_output(x):
+    return x.stdout.decode("utf-8").rstrip().split("\n")
+
+
+def submodule_status():
+    res = run(["git", "submodule", "status"], stdout=PIPE, check=True)
+    lines = subprocess_output(res)
+    parsed = [[x[0] != " "] + x[1:].strip().split(" ")[:2] for x in lines]
```
EmmaLRussell: This is because you get a blank space at the start of the submodule status line if it matches the repository and some  character like + or - if it doesn't, so you can tell if local has moved on? Some light commenting for dullards like me might be nice here!

scripts/create-latest
```diff
@@ -0,0 +1,115 @@
...
+
+
+def update_latest():
+    remote_status = get_remote_status()
+    print("Testing if any submodules need updating...")
+    res = run(UPDATE, stdout=PIPE, check=True)
+    message = "Auto: Update versions to latest\n\n" + \
+              "\n".join(subprocess_output(res))
+    local_status = get_local_status()
+    if remote_status["sha_parent"] != local_status["sha"]:
```
EmmaLRussell: So you expect the parent of the remote origin/latest to be same commit as the local master if nothing has changed? Because latest always branches afresh from the latest master, so if these are different, it means master has moved on..

scripts/create-latest
```diff
@@ -0,0 +1,115 @@
...
+    else:
+        print("\n** {}/{} is up to date ** \n".format(ORIGIN, LATEST))
+        new_build = False
+    if new_build:
+        print("Creating new {}...".format(LATEST))
+        run(["git", "branch", "-D", LATEST], check=False)
+        run(["git", "checkout", "-b", LATEST], check=True)
+        if local_status["changed"]:
+            print("Creating commit with submodule changes...")
+            run(["git", "add"] + local_status["changed"], check=True)
```
EmmaLRussell: OK, so adding all the paths to the submodules that have changed, that originally came from 'git submodules status'. Got confused about that parsing for a while but think I understand it now!

scripts/create-latest
```diff
@@ -0,0 +1,115 @@
...
+            run(["git", "commit", "-m", message], check=True, stdout=DEVNULL)
+        print("Updating remote...")
+        url = "git@github.com:vimc/montagu"
+        run(["git", "remote", "set-url", ORIGIN, url], check=True)
```
EmmaLRussell: Why do we need to set the url?

scripts/create-latest
```diff
@@ -0,0 +1,115 @@
...
+    sha = subprocess_output(
+        run(["git", "rev-parse", MASTER], stdout=PIPE, check=True))[0]
+    versions, changed = submodule_status()
+    return {"sha": sha, "versions": versions, "changed": changed}
+
+
+def update_latest():
+    remote_status = get_remote_status()
+    print("Testing if any submodules need updating...")
+    res = run(UPDATE, stdout=PIPE, check=True)
```
EmmaLRussell: Still not sure where this UPDATE script actually is, but assuming it's the thing that's going to update the local submodules to their actual latest commits....

scripts/create-latest
```diff
@@ -0,0 +1,115 @@
...
+
+import os
+from subprocess import run, DEVNULL, PIPE
+
+# We *probably* won't change any of these, but at least it means we
+# could if we wanted to.
+LATEST = "latest"
+ORIGIN = "origin"
+MASTER = "master"
+UPDATE = "./src/update_versions_to_latest_master.py"
```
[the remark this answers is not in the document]
richfitz: oops

scripts/create-latest
```diff
@@ -0,0 +1,115 @@
...
+    lines = subprocess_output(res)
+    parsed = [[x[0] != " "] + x[1:].strip().split(" ")[:2] for x in lines]
+    versions = {x[2]: x[1] for x in parsed}
+    changed = [x[2] for x in parsed if x[0]]
+    return [versions, changed]
+
+
+def get_remote_status():
+    ref = "{}/{}".format(ORIGIN, LATEST)
+    res = run(["git", "rev-parse", ref, "--"], stdout=PIPE, stderr=DEVNULL)
```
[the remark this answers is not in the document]
richfitz: The `--` is necessary for git to disambiguate references from paths. Because the reference is on the lhs of the `--` git knows its a reference even when it does not exist

scripts/create-latest
```diff
@@ -0,0 +1,115 @@
...
+        run(["git", "remote", "set-url", ORIGIN, url], check=True)
+        run(["git", "push", "--force", ORIGIN, LATEST], check=True)
+    else:
+        print("Restoring previous {}".format(MASTER))
+        run(["git", "submodule", "update", "--recursive"], stdout=DEVNULL)
+
+
+def preflight():
+    print("checking working directory")
+    if not os.path.isfile(UPDATE):
```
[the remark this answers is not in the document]
richfitz: This is checking our working directory is montagu's root

PR #141, opened 2019-01-18
i2434 - don't allow import user to write to enum tables

littledice:
Tested locally that this does in fact prevent the import user from writing to the specified tables.


src/database.py
```diff
@@ -16,10 +16,10 @@
 def user_configs(password_group):
     # Later, read these from a yml file?
     return [
-        UserConfig(api_db_user, 'all', VaultPassword(password_group, api_db_user)),
-        UserConfig('import', 'all', VaultPassword(password_group, 'import')),
-        UserConfig('orderly', 'all', VaultPassword(password_group, 'orderly')),
-        UserConfig('readonly', 'readonly', VaultPassword(password_group, 'readonly')),
+        UserConfig(api_db_user, 'all', [], VaultPassword(password_group, api_db_user)),
+        UserConfig('import', 'all', ["gavi_support_level", "activity_type"], VaultPassword(password_group, 'import')),
```
richfitz: what is the thinking behind letting anyone write to these tables?  if this was set for all users then
* root, who runs the migrations, can still edit them
* nobody else can
* we don't need the logic here

richfitz: there are other tables too that might be worth adding, e.g. `burden_estimate_set_status`, `touchstone_status`, `impact_outcome`, `responsibility_set_status`, `burden_outcome`.  See `grep -i "insert into" migrations/sql/*.sql` in `montagu-db` for a list of tables that the migration looks after

littledice: Both good points!

PR #143, opened 2019-01-22
i2516: More relaxed orderly api server configuration

richfitz:
Can we test this with your new vagrant config Emma? Using any unknown server name should have the same effect as previously using teamcity now

### DIFF
--- a/src/orderly.py
+++ b/src/orderly.py
@@ -34,12 +34,19 @@ def configure_orderly(service, initialise_volume):
 def configure_orderly_envir(service):
     password_group = service.settings['password_group']
     api_server = service.settings['instance_name'].lower()
+    # These are the entries in montagu-reports/orderly_config.yml:
+    # update here if more are added.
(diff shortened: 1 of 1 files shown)


src/orderly.py
```diff
@@ -34,12 +34,19 @@ def configure_orderly(service, initialise_volume):
...
+    # update here if more are added.
+    known = ["science", "uat", "production"]
+    if api_server in known:
+        slack_url = '"{}"'.format(
+            get_secret("slack/orderly-webhook", field="url"))
+    else:
+        api_server = "~"
+        slack_url = "~"
+
     if api_server == "(unknown)" or api_server == "teamcity":
```
EmmaLRussell: Shouldn't need this bit anymore, if we've already set to ~ for anything other than known?

richfitz: thanks, totally should have been removed

richfitz: removed now

PR #144, opened 2019-01-23
i2537: Update instance names in settings

richfitz:
Also deletes the restore-test.json settings that was no longer
used

I see a build failure but I'm pretty sure this is not my fault?

http://teamcity.montagu.dide.ic.ac.uk:8111/viewLog.html?buildId=28809&tab=buildResultsDiv&buildTypeId=montagu_deploy


littledice:
That's annoying, looks like muting the test in TC isn't enough in this case. @EmmaLRussell do you mind adding a junit "ignore" decorator to the tests in question in the API code? That'll shut them up properly.

richfitz:
I think that the issue is that the exit code from the test runner is 1?

PR #146, opened 2019-02-06
I2606 - add caddy server to montagu

littledice:
* Added Caddy server to docker network
* Logic to copy report artefacts from orderly volume to static volume
* Point API to the branch which issues the new cookies for Caddy

This is currently deployed to science for testing:

Visiting https://support.montagu.dide.ic.ac.uk:11443/model-review/ should give a 401 if not logged in, a 403 if logged in as a user without one of the admin/funder/developer roles, and content should be viewable if logged in as a user with one of the above roles. Individual users have access to individual sub-paths, but that is going to be a bit harder to test!


docker-compose.yml
```diff
@@ -86,8 +86,15 @@ services:
     volumes:
       - orderly_volume:/orderly
     command: --port 8321 --go-signal /orderly_go /orderly
+  static:
+    image: ${MONTAGU_REGISTRY}/montagu-static:${MONTAGU_STATIC_VERSION}
+    restart: always
+    logging: *log-journald
+    volumes:
+      - static_volume:/www
```
richfitz: If we wanted to persist logs, we might add

```
      - static_logs:/var/log/caddy
```

and 
* add an extra volume below in `volumes:` 
* update the paths here https://github.com/vimc/montagu-static/blob/8c9e0e06a8808dd2831f85a2c716ffeaa43eabc1/caddy/Caddyfile#L6-L7 to add the `caddy` part to the path

Not sure how worthwhile this would be, but I think we increasingly do this on other bits of dockerised infrastructure

src/model_review_2019
```diff
@@ -0,0 +1,38 @@
+native-diagnostics-burden-report-drafts/20190205-151702-1ba5e47a/*,PHE-Vynnycky
```
richfitz: should this file have an `.csv` ending since that seems to be the format?

richfitz: Can we move this into a directory (e.g. `static/model-review/2019.csv`) - and update the configure there so that we could just add more bits that need copying by adding new csv files.  The pattern could be that a csv at `static/<path>/<file>.csv` would copy files to paths within `static/<path>/<file>`

richfitz: If it is possible to lift the `static` path out of `src` and put it at the montagu root (like submodules) that would be preferable IMO - I think we have a `montagu_root` variable or something to help with this.  It's not really source as such.  And can you add a small `README.md` describing the approach to the `static` directory?

littledice: I changed the logic as suggested so that `static/model-review/2019/config.csv` would copy listed files to `www/model-review/2019/` on the static server. But now we're kind of mixing ways of doing things because the config format is `file_glob, destination_folder`, so the full destination path is specified both by the config file location, and the config itself. As I'm updating the README this feels confusing? 

To make consistent we could either break down the config file so we have
`./static/model-review/2019/<group-name>/config` for each group, where each config file now just lists file globs and the destination is fully determined by the location of the config file. Or we could stick with a single config file, `static/config_model_review_2019.csv` but list the full destination is specified in the file itself, where a row looks like 
```
native-diagnostics-burden-report-drafts/20190131-123847-53fe189e/*,model-review/2019/IC-Hallett
```

richfitz: What about treating the path in the csv as relative to the location of the csv?

So

```
static/model-review/2019/config.csv
```
and
```
native-diagnostics-burden-report-drafts/20190131-123847-53fe189e/*,IC-Hallett
```

littledice: That's what I have done here. So you think this mixed approach is fine/not confusing?

littledice: It seemed ok to me until I started writing the README, and then it seemed weird!

richfitz: sorry, trying to multitask too much :confused: 

I don't personally find that confusing, but if you do someone else might.

But if we did another set of static files that were much simpler we might end up with

```
static/summary-2019/config.csv
```

and

```
summary-2019/20190131-123847-53fe189e/*,.
```

for example, which would seem ok to me

littledice: Maybe @EmmaLRussell could weigh in on this, here's my explanation in the README, does this seem sensible/understandable? https://github.com/vimc/montagu/pull/146/commits/ab95da24d605bdbd4001cdb26cb1381225e39ffe

EmmaLRussell: OK, I think I see both your points. I agree with Alex that it does seem a bit weird to be specifying the path both through the location of the config file and its content (paths relative to the file location). But then breaking down the config file into lots of files in different locations seems like it would probably be annoying to maintain. So I think where you've got to is fine as long as it's clear through the readme (which it is!)

littledice: Great, thanks! This does seem like the least annoying way to write things, so lets go with it :smile:

PR #147, opened 2019-02-08
I2616 - just host model documentation

littledice:
Also:
https://github.com/vimc/montagu-static/pull/3
and
https://github.com/vimc/montagu-proxy/pull/19

### DIFF
--- a/src/service.py
+++ b/src/service.py
@@ -157,9 +157,7 @@ class MontaguService:
             self.stop_metrics()
         except Exception as e:
             print("Error when stopping Metrics container: {}".format(str(e)))
-        print("Wiping static file volume")
-        static_volume = self.client.volumes.get(self.volumes["static"])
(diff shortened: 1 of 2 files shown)


src/service.py
```diff
@@ -176,6 +176,12 @@ def stop(self):
                 print("Killing orderly container failed - continuing")
                 pass
         compose.stop(self.settings)
+        print("Wiping static file volume")
```
littledice: Occurred to me that we also need to delete files from the volume that are not in the config, hence wiping this volume each time. Kind of means that a volume is the wrong technology for this, if we don't want it to persist, but copying files into the container would be a faff.

richfitz: The alternative way would be a new volume utility for clearing the volume just before doing a copy (would be the same lightweight container approach running `rm -rf /volume/*`) but this way seems no less complicated

PR #152, opened 2019-02-26
I2707 - add all global roles to test user on deploy

hillalex:
Depends on https://github.com/vimc/montagu-api/pull/368


hillalex:
Changed my mind about this.

PR #157, opened 2019-06-21
Vimc 2958 Give test user access to all touchstones, and add test modeller user

EmmaLRussell:


src/cli.py
```diff
@@ -61,19 +61,22 @@ def add_test_user():
...
-    run(command + [name] + args)
-
-    args = ["addRole", "test.user", "user"]
 
-    run(command + [name] + args)
+    run_cmd(command, name, ["add", "Test User", "test.user", "test.user@imperial.ac.uk", "password", "--if-not-exists"])
+    run_cmd(command, name, ["addRole", "test.user", "user"])
+    run_cmd(command, name, ["addRole", "test.user", "reports-reviewer"])
+    run_cmd(command, name, ["addRole", "test.user", "touchstone-reviewer"])
+    run_cmd(command, name, ["addUserToGroup", "test.user", "ALL"])
```
hillalex: So this user is supposed to be the admin user? Shall we make this clearer and call them "admin.user"? And I guess they don't want to be added to all modelling groups (in fact I think this is what causes them to have an impossibly long token)

src/cli.py
```diff
@@ -61,19 +61,22 @@ def add_test_user():
...
 
-    args = ["addRole", "test.user", "reports-reviewer"]
-
-    run(command + [name] + args)
+    run_cmd(command, name, ["add", "Test Modeller", "test.modeller", "test.modeller@imperial.ac.uk", "password", "--if-not-exists"])
+    run_cmd(command, name, ["addRole", "test.modeller", "user"])
+    run_cmd(command, name, ["addRole", "test.modeller", "reports-reader"])
+    run_cmd(command, name, ["addRole", "test.modeller", "member", "modelling-group:IC-Garske"])
+    run_cmd(command, name, ["addRole", "test.modeller", "member", "modelling-group:Harvard-Sweet"])
+    run_cmd(command, name, ["addUserToGroup", "test.modeller", "ALL"])
```
hillalex: This line supersedes the last two. But probably best if they're just a member of the 2 selected groups?

src/cli.py
EmmaLRussell: Renamed to test.admin.

hillalex:
This looks good to me, but I guess we should fix the build first and try running again (looks like some other orderly based issue?) http://teamcity.montagu.dide.ic.ac.uk:8111/viewLog.html?buildId=49421&tab=buildResultsDiv&buildTypeId=montagu_deploy

PR #158, opened 2019-06-24
VIMC-2947 support switch from reporting portal to orderly web

hillalex:
Updates the proxy which will now point to the version that proxies OW. Changed the "stop" logic to stop containers rather than bringing down the network.

### DIFF
--- a/src/compose.py
+++ b/src/compose.py
@@ -11,6 +11,11 @@ def start(settings):
 def stop(settings):
     command = "stop" if settings["persist_data"] else "rm --force --stop"
     run(command, settings)
+    p = Popen(["docker", "system", "prune"], env=get_env(settings), shell=True)
+    p.wait()
+    if p.returncode != 0:
+        raise Exception("An error occurred: docker returned {}".format(
+            p.returncode))
 
 
 def pull(settings):
--- a/src/compose.py
+++ b/src/compose.py
@@ -9,7 +9,7 @@ def start(settings):
 
 
 def stop(settings):
-    command = "down" if settings["persist_data"] else "down --volumes"
+    command = "stop" if settings["persist_data"] else "rm --force --stop"
     run(command, settings)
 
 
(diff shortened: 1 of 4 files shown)


src/compose.py
```diff
@@ -9,8 +9,13 @@ def start(settings):
 
 
 def stop(settings):
-    command = "down" if settings["persist_data"] else "down --volumes"
```
hillalex: `down` brings down the network which we don't want - we just want to stop the images. If `persist_data` is true we can also `rm` the images, meaning when we run `system prune` the volumes will also be cleaned up.

src/service.py
```diff
@@ -161,20 +161,8 @@ def stop(self):
         print("Stopping Montagu...({}: {})".format(
             self.settings["instance_name"], self.settings["docker_prefix"]),
               flush=True)
-
-        # As documented in VIMC-805, the orderly container will
```
richfitz: I've moved this into https://vimc.myjetbrains.com/youtrack/issue/mrc-351 as we don't currently support it in orderly-web-deploy and it would be good to have

src/compose.py
```diff
@@ -9,8 +9,13 @@ def start(settings):
...
-    run(command, settings)
+    run("stop", settings)
```
richfitz: The other behaviour difference we'll see here with `stop` instead of `down` is that the containers won't be removed.  Does that end up being ok with restarts?

src/compose.py
```diff
@@ -9,8 +9,13 @@ def start(settings):
...
+    if not settings["persist_data"]:
+        p = Popen("docker system prune -a --volumes -f", shell=True)
```
richfitz: I think this should limit itself to the montagu containers, networks and volumes only otherwise it could end up accidentally wiping another volume

src/compose.py
hillalex: It does seem to. But maybe we should in fact remove them.

src/compose.py
hillalex: Oh yeah, good point (!)

PR #160, opened 2019-07-03
MRC-294: remove orderly, reporting api, reporting portal

richfitz:
This PR will remove orderly, the reporting api and the reporting portal from the montagu deployment.

There's still a little bit of a tangle because we need to copy some files from the orderly volume, but that can be left as-is for now I believe.


src/service_config/static_server_config.py
```diff
@@ -58,8 +58,7 @@ def get_artefacts(path):
 
 
 def configure_static_ssh(service):
-    needs_ssh = service.settings['clone_reports'] or \
-                service.settings['initial_data_source'] == 'restore'
+    needs_ssh = service.settings['initial_data_source'] == 'restore'
```
hillalex: or `copy_static_files`

hillalex:
Do you mind rolling in 1 change (small bug I introduced in the last PR) - move this line https://github.com/vimc/montagu/blob/master/src/service.py#L165 into a try except block? If you manually run `./stop.py` and then `deploy.py` it errors because it tries to remove a container that has already been removed.

PR #161, opened 2019-08-15
VIMC-2986 support notifying Slack when estimates uploaded

hillalex:
Inject webhook from vault into the api config. I have already written the value into the vault.


hillalex:
> Looks good! Can we test this on uat by setting password group in the deploy settings?

Hmm, I *think* this would be ok. @richfitz can you think of any bad consequences of temporarily deploying UAT with production passwords?

richfitz:
No should be fine

PR #163, opened 2020-01-20
vimc-3230 Added orderly web api url

EmmaLRussell:


settings/dev.json
```diff
@@ -18,5 +18,6 @@
   "update_on_deploy": false,
   "include_guidance_reports": false,
   "use_production_db_config": false,
-  "copy_static_files": false
+  "copy_static_files": false,
+  "orderly_web_api_url": "http://localhost:8443/reports/api/v1"
```
hillalex: How come this is different from the suggested value for local dev? https://github.com/vimc/montagu/pull/163/files#diff-7f43a41e5e875b219434f43a6fb8437dR72

EmmaLRussell: Good question!

settings/teamcity.json
```diff
@@ -18,5 +18,6 @@
     "update_on_deploy": false,
     "copy_static_files": false,
     "include_guidance_reports": false,
-    "use_production_db_config": false
+    "use_production_db_config": false,
+    "orderly_web_api_url": "http://localhost:8443/reports/api/v1"
```
hillalex: This 1 should be the same as dev presumably?

hillalex:
I pointed the api and webapps submodules to the latest commits and the create user tests are failing, but I think just because of the wrong api url in the teamcity settings http://teamcity.montagu.dide.ic.ac.uk:8111/viewLog.html?buildId=78299&tab=buildResultsDiv&buildTypeId=montagu_deploy

src/test.py
```diff
@@ -65,6 +66,55 @@ def work():
...
+            "docker", "run", "--rm",
+            "-v", "demo:/orderly",
+            ow_migrate_image
+        ], check=True)
+
+        ow_cli_image = get_image_name("orderly-web-user-cli", "master")
+        pull(ow_cli_image)
+
+        #user for api blackbox tests
+        run([
```
hillalex: Can we pull these repeated blocks out into their own functions, `add_user` and `grant_permissions`?

PR #164, opened 2020-02-11
VIMC-3421 remove annex user mappings

hillalex:
Removes the setting up of users on the main db that have access to the annex via the foreign data wrapper. Also points the db and api submodules to corresponding branches:
https://github.com/vimc/montagu-api/pull/419
https://github.com/vimc/montagu-db/pull/139

All tests passing here indicate that the removal hasn't broken anything as far as the api and webapps go, but I'm not sure about orderly or any other programs that talk to the database.


hillalex:
> Is the idea that annex will get its own deployment script/project, and the setup part for annex in database.py will be removed after that? Or does it remain part of the montagu deploy?

I presume it stays part of Montagu. But we have some other mechanism for schema changes so that the science team can carry them out with less friction, maybe via data imports.

hillalex:
> > Is the idea that annex will get its own deployment script/project, and the setup part for annex in database.py will be removed after that? Or does it remain part of the montagu deploy?
> 
> I presume it stays part of Montagu. But we have some other mechanism for schema changes so that the science team can carry them out with less friction, maybe via data imports.

As pointed out in standup it doesn't even get deployed by this tool at the moment! :woman_facepalming: Having said that, getting users onto the annex, pulling credentials from vault etc, all happenes here currently. So I don't want to remove that until we have a replacement in place - @richfitz did you say you have a deploy script for the annex incoming? Does it handle users?

PR #170, opened 2020-07-24
vimc-4010 Add montagu-task-queue to deployment

EmmaLRussell:
This adds the montagu-task-queue and associated message queue containers to the montagu network, and does some basic configuration, by reading out the config.yml from the container, overwriting settings where required and saving back into the container. 

Includes a configuration test that a test task can be run. 

Still to do in future branches:
- Actually create the montagu user when required if it does not already exist (and pull a password from the vault?)
- Configure with 'real' diagnostic reports config to run on prod/science/uat. 
- Add fake smtp containers for uat and science.


src/service_config/task_queue_config.py
```diff
@@ -0,0 +1,43 @@
...
+    if montagu_user is not None:
+        montagu["user"] = montagu_user
+        # TODO: password
+
+    # Task queue needs orderly-web url without /api/v1 suffix
+    ow_url_trimmed = orderly_web_url.replace("/api/v1", "")
+    config["servers"]["orderlyweb"]["url"] = ow_url_trimmed
+
+    if is_prod:
+        smtp = config["servers"]["smtp"]
```
EmmaLRussell, line 31: This is just copied over from montagu api config, but that has a user name and password for smtp too so I guess we'll need to add these into the task queue config too for this to work on production.

container_config/task_queue/real_diagnostic_reports.yml
```diff
@@ -0,0 +1 @@
+tbd:
```
EmmaLRussell, line 1: The real diagnostic task config will be added in another ticket

docker-compose.yml
```diff
@@ -37,7 +37,7 @@ services:
     volumes:
       - db_volume:/pgdata
   contrib:
-    image: vimc/montagu-contrib-portal:${MONTAGU_CONTRIB_PORTAL_VERSION}
+    image: ${VIMC_REGISTRY}/montagu-contrib-portal:${MONTAGU_CONTRIB_PORTAL_VERSION}
```
richfitz, line 40: Should this *always* be `vimc` now, as we move to deprecate the old registry?

src/deploy.py
```diff
@@ -129,6 +129,14 @@ def configure_montagu(service, data_exists):
     configure_api(service, passwords['api'], token_keypair_paths,
                   service.settings["hostname"], is_prod,
                   service.settings["orderly_web_api_url"])
+
+    task_queue_user = "MONTAGU_TASK_QUEUE" if service.settings["use_real_diagnostic_reports"] else "test.user@example.com"
+    # TODO: Add the user and password if does not exist and grant required perms (run, publish)
```
richfitz, line 134: Is there a ticket for this, or were you planning on doing it here?

container_config/orderlyweb/config.properties
```diff
@@ -1 +1,2 @@
-montagu.api_url=http://montagu_api_1:8080/v1
\ No newline at end of file
+montagu.api_url=http://montagu_api_1:8080/v1
```
richfitz, line 1: commenting here so that it's nearby, but why are the go_signal files included? You are using `touch` to create them in the startup. I feel like I am not seeing something in the setup or testing here where these sneak across, and that could do with a comment if it's the case

EmmaLRussell: Yeah, this is actually just used in the test script I think but could probably do without it and use touch as elsewhere, will put that in. :+1:

src/deploy.py
EmmaLRussell: There's a ticket: https://mrc-ide.myjetbrains.com/youtrack/issue/VIMC-4218

docker-compose.yml
EmmaLRussell: It should be, not sure all the images are getting uploaded there yet - we've got a couple of epics in the current sprint for this we should try to move forward...

EmmaLRussell:
This branch has now been successfully deployed to uat - new containers are there, but not configured to actually do anything yet!

docker-compose.yml
hillalex: The portals are on docker hub and in fact no longer exist on the public registry, hence 'vimc' here!

hillalex:
> This branch has now been successfully deployed to uat - new containers are there, but not configured to actually do anything yet!

I'm suprised this has worked because the admin and contrib portal images should no longer exist on the local registry? Unless the versions of those pinned here are old...

EmmaLRussell:
Oh I possibly didn't update all the submodules except for the new one. But this branch is using vimc registry for the portals, it's just I added a variable for it, so it should work anyway I think?

> 
> I'm suprised this has worked because the admin and contrib portal images should no longer exist on the local registry? Unless the versions of those pinned here are old... Seems likely since this branch has been around a while. I suggest doing a merge from master to update those.

hillalex:
> Oh I possibly didn't update all the submodules except for the new one. But this branch is using vimc registry for the portals, it's just I added a variable for it, so it should work anyway I think?
> 
> > I'm suprised this has worked because the admin and contrib portal images should no longer exist on the local registry? Unless the versions of those pinned here are old... Seems likely since this branch has been around a while. I suggest doing a merge from master to update those.

Oh my bad, that's totally correct then! I misinterpreted Rich's comment and just assumed VIMC_REGISTRY meant the local reg.

PR #174, opened 2020-08-18
vimc-4219 Add fake smtp server to non-Production environments

EmmaLRussell:
Adds a fake smtp container, which is conditionally included from its own compose file if setting `fake_smtp` is true.


src/compose.py
```diff
@@ -19,7 +19,8 @@ def pull(settings):
 
 def run(args, settings):
     docker_prefix = settings["docker_prefix"]
-    prefix = 'docker-compose --project-name {} '.format(docker_prefix)
+    staging_file =  "-f ../docker-compose.staging.yml" if settings["fake_smtp"] else ""
```
EmmaLRussell, line 22: This seems to be the way to conditionally include containers using compose, by having them in separate file which yoi spin up together.

src/test.py
```diff
@@ -69,12 +71,21 @@ def work():
...
+        print("Running task queue integration tests")
         app = celery.Celery(broker="pyamqp://guest@localhost//", backend="rpc://")
         sig = "src.task_run_diagnostic_reports.run_diagnostic_reports"
         signature = app.signature(sig, ["testGroup", "testDisease"])
         versions = signature.delay().get()
         assert len(versions) == 1
         assert len(versions[0]) == 24
+        # check expected notification email was sent to fake smtp server
+        emailstr = run([
+            "docker", "exec", "montagu_task-queue_1", "curl", "http://montagu_fake_smtp_server_1:1080/api/emails"
```
EmmaLRussell, line 83: Testing directly from localhost doesn't work on TeamCity so fetching from within container here.

hillalex (approved):

PR #179, opened 2020-08-20
vimc-4218 Generate password and add user for task queue

EmmaLRussell:


src/cli.py
```diff
@@ -30,53 +32,63 @@ def add_secure_config(password_group):
...
+
+    run_cmd(command, image, ["add", name, id, email, password, "--if-not-exists"])
+
+
+def get_network():
+    settings = get_settings(quiet=True)
+    service = MontaguService(settings)
+    return service.network_name
+
+def docker_run_cmd(network):
```
hillalex, line 83: I took "run_cmd" to be a verb, maybe "get_docker_run_cmd" might avoid confusion?

hillalex (commented):
This all makes sense to me. Tiniest suggestion re naming - happy for you to merge after that.

Re hiding the user in admin screens, for sure that seems like a good idea in Montagu, I'm ever so slightly reluctant to add more Montagu specific logic into OW (although maybe I shouldn't be, given what a massive part of the auth architecture it already is!) I guess one way of doing it in quite an abstract way would be to add a flag to the orderlyweb_user table like `editable` or something.

EmmaLRussell:
Hm, yeah, good point.

> Re hiding the user in admin screens, for sure that seems like a good idea in Montagu, I'm ever so slightly reluctant to add more Montagu specific logic into OW (although maybe I shouldn't be, given what a massive part of the auth architecture it already is!) I guess one way of doing it in quite an abstract way would be to add a flag to the orderlyweb_user table like `editable` or something.

PR #182, opened 2020-09-22
updated modeller emails and group names

kimwoodruff:
Hi Alex and Emma - I've updated the emails in the PI emails section. I've also added 3 new groups (IVI-Kim, JHU-Lee and Yale-Pitzer) and emails for them. Could you check my syntax is OK? thanks!


hillalex (approved):
Perfect!

PR #183, opened 2020-09-23
added Imperial test user group and email

thuyvuong1:
added "Imperial-test user" and email for the PI (t.vuong@imperial.ac.uk)

### DIFF
--- a/src/generate_real_diagnostic_reports_config.py
+++ b/src/generate_real_diagnostic_reports_config.py
@@ -45,8 +45,8 @@ def generate():
         "PSU-Ferrari-WHO": ["Measles"],
         "UND-Moore": ["JE"],
         "UND-Perkins": ["YF"],
-        "Yale-Pitzer": ["Typhoid"]
-        "Imperial-test user" [    ]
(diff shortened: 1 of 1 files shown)


kimwoodruff (changes requested):
Hi Thuy - great work on your 1st pull request! Could you make these small changes? 
1) change "Imperial-test user" to "VIMC". if you're wondering "how could I have known that?", it's not obvious, but each group has a long name (which appears in the grey horizontal bar in the contribution portal), and a shorter group-ID (which appears in the URL). We need to use the group-ID here. If you try adding yourself to a few modelling groups and explore the grey bars and URLs in the contribution portal, you should see what I mean. 
2) add a final comma on line 77 and delete it on line 78. (I'm not sure how crucial this syntax is to be honest, but it's good practice to be consistent.) 
3) add typhoid and cholera in the square brackets on line 49. (If you look in the contribution portal, you'll see that the Imperial-test user group is set up for these 2 diseases.) 
Thanks!

kimwoodruff (approved):

PR #186, opened 2020-10-07
vimc-4297: Add buildkite support

richfitz:
This PR adds basic buildkite support. Resisting the urge to refactor much here!

Becuase of the worker change possibility there's no real point running "clear docker" as a step - instead that's run within the ci script. The settings check script can be run concurrently.


src/teamcity.py
```diff
@@ -1,106 +0,0 @@
-from getpass import getpass
```
hillalex, line 1: what on earth was all this?

src/test.py
```diff
@@ -25,12 +25,9 @@
...
 
 
-def run_in_teamcity_block(name, work):
-    print("##teamcity[blockOpened name='{name}']".format(name=name))
-    try:
-        work()
-    finally:
-        print("##teamcity[blockClosed name='{name}']".format(name=name))
+def run_in_buildkite_block(name, work):
+    print("--- " + name)
```
hillalex, line 29: I didn't know you could do this to get buildkite to recognise different blocks! nice

hillalex (commented):
This is great, thank you! One minor thing which would be nice to fix, is that the `api_blackbox_tests` task doesn't show up nicely in the build log seemingly because the formatted output from the cli in the previous task doesn't force a new line: https://buildkite.com/mrc-ide/montagu/builds/11#b0984e43-a3e4-4b3f-a174-1fc6f0412afe/402-889

Would be helpful to fix that, maybe just by adding an extra `print("")` before  `print("--- " + name)` so that the blackbox tests task is easy to find in the log?

src/teamcity.py
richfitz: not entirely sure, something to do with fetching artifacts that was not being currently used.

hillalex (approved):

PR #191, opened 2020-11-10
vimc-4392 Create redis and flower containers for task-queue's message queue

EmmaLRussell:
Currently targeting this branch of the task-queue submodule: https://github.com/vimc/montagu-task-queue/pull/19
And this branch of the api: https://github.com/vimc/montagu-api/pull/447

Seemed to need to update the other submodules too - they was getting updated automatically on teamcity before?

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -78,6 +78,14 @@ services:
       - "6379:6379"
     volumes:
       - mq:/data
+  flower:
+    image: mher/flower:0.9.5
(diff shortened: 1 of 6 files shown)


hillalex (approved):
Updating submodules has always been manual. We just don't do it very often these days as we're not often shipping new code to Montagu! In this case I guess it's all the upload coverage stuff, but that's fine.

PR #192, opened 2020-12-10
vimc-4459 Run copy-vis-tool.sh on deploy

EmmaLRussell:
Run the copy script at the end of deploy, and report error if it fails. I haven't added a test, as it will always fail in test environments which do not have the specified report. 

I have run it successfully on UAT, and files were copied and available at the /2020/datavis endpoint.


EmmaLRussell:
Actually, might be best not to merge this one now, as we're likely to have at least one more montagu deploy where we don't want the public tool link to go live yet..

richfitz (approved):
ok, I will leave this approved though as I'd apparently looked through but failed to actually tick the box

PR #193, opened 2020-12-16
vimc-4475 Remove use of private docker registry from release scripts

EmmaLRussell:
Apparently we haven't done a proper montagu release since we removed the private docker registry. This branch simplifies the `tag-images` script by removing the option to publish to the vimc registry - because that's what it always uses now. Also updates instructions to user.


richfitz (approved):
sorry, looks like I had not merged this!

PR #201, opened 2021-03-12
VIMC-4567 Migrate burden estimate upload feed to Teams

mwoodbri:
Required for vimc/montagu-api#457

URL has been added to Vault under the relevant path. It includes auth token.

Slack appears to also be used for notifications of new deploys(?) but that uses a different secret


EmmaLRussell (approved):
Looks good to me. Guess we can't test the whole thing in anger without deploying to prod!

PR #208, opened 2021-07-22
VIMC-4798 Update release script to latest YouTrack API

mwoodbri:
This has been tested by creating [v2.32.0](https://mrc-ide.myjetbrains.com/youtrack/issues?q=Fixed%20in%20build:%20v2.32.0)

New API is rather less ergonomic than its predecessor...

### DIFF
--- a/scripts/release/tickets.py
+++ b/scripts/release/tickets.py
@@ -25,23 +25,28 @@ https://www.jetbrains.com/help/youtrack/standalone/Manage-Permanent-Token.html
 
 class Ticket:
     def __init__(self, data):
-        self.id = data["id"]
-        self.fields = {}
-        for field in data["field"]:
-            self.fields[field["name"]] = field["value"]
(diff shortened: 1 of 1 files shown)


scripts/release/tickets.py
```diff
@@ -25,23 +25,28 @@ def get_token():
...
+            for customField in data["customFields"]
+        )
+        self.fields["summary"] = data["summary"]
 
     def get(self, field):
         return self.fields[field]
 
     def state(self):
-        return ", ".join(self.get("State"))
+        return self.get("State")
```
mwoodbri, line 42: The new API only returns a single value for each `State` field. I guess multiple `State` fields could be returned but I can't see how it is possible to set this in the UI, and I can't find any issues with more than one value.

EmmaLRussell (approved):
I see what you mean about the new API!

PR #212, opened 2021-08-31
vimc-4841_regen Regenerate config file from updated script

EmmaLRussell:
Regenerate the config file from the script changes made in https://github.com/vimc/montagu/pull/211


r-ash (approved):
Thanks for this!

PR #220, opened 2021-11-29
configure task q to create yt tickets

hillalex:
Generates new task q config with science team assignees for each group, adds a test for the YT integration, and points the  task q submodule to the latest changes.

~~Will require a `pip3 install --user -r src/requirements.txt` when re-deploying as adds a new python dependency, the YouTrack api client.~~  Actually this is only a test dependency, so shouldn't matter for a deploy and the CI agents do this already


src/test.py
```diff
@@ -74,11 +76,13 @@ def work():
 def task_queue_integration_tests():
     def work():
         print("Running task queue integration tests")
+        yt = YTClient('https://mrc-ide.myjetbrains.com/youtrack/',
+                      token=os.environ["YOUTRACK_TOKEN"])
```
hillalex, line 80: this has been added to the build agents here: https://github.com/reside-ic/reside-hyperv-scripts/pull/20

src/service_config/task_queue_config.py
```diff
@@ -52,6 +53,9 @@ def configure_task_queue(service,
         smtp["user"] = "montagu"
         smtp["password"] = get_secret("email/password")
 
+    if is_prod:
+        config["servers"]["youtrack"]["token"] = \
+            get_secret("vimc-robot/youtrack-task-queue-token")
```
hillalex, line 58: In the absence of a provided config value (e.g. in the test context on BuildKite) the task q will look for an environment variable called "YOUTRACK_TOKEN" (see https://github.com/vimc/montagu-task-queue/pull/24/files#diff-f81481734ef4384fb31e9f3cb8d3afacf27470dedc4210533a3fc4747419c33dR34) which now exists on the BuildKite agents.

richfitz (approved):

PR #221, opened 2022-04-09
vimc-6327 Add burden estimate files volume

EmmaLRussell:
This branch:
- adds a volume to docker compose, `burden_estimate_files`, which is used by the api and task queue containers 
- Configures the task queue to use a minimum file age of 1 hour when archiving files. 
- Opens the permissions to the volume when the api starts, so that it is accessible to the task queue worker user

Deployed to UAT, or you can test locally with `src/deploy.py`. 

You can run the task with:
`curl -X POST -d '{"args":["burden_estimate_files"]}' http://localhost:5555/api/task/send-task/archive_folder_contents`

Any files in the volume (at the top level) which are older than 1 hour should be deleted. I've left a file, `test.txt`, in the volume on UAT, to make this easy to test, or you could edit the config file in the task queue container to drop the minimum file age. 

To list files in the volume:
`docker exec montagu_api_1 ls tmp/burden_estimate_files`

To add a file to the volume:
`docker exec montagu_api_1 touch tmp/burden_estimate_files/test.txt`

### DIFF
--- a/src/service_config/api_config.py
+++ b/src/service_config/api_config.py
@@ -25,9 +25,6 @@ def configure_api(service, db_password: str, keypair_paths, hostname, is_prod: b
     generate_api_config_file(service, config_path, db_password, hostname,
                              is_prod, orderly_web_api_url, celery_flower_host)
 
-    print("- Making burden estimate files accessible to task queue")
-    service.api.exec_run("chmod a+w tmp/burden_estimate_files")
-
     print("- Sending go signal to API")
     service.api.exec_run("touch {}/go_signal".format(config_path))
 
--- a/src/service_config/api_config.py
+++ b/src/service_config/api_config.py
@@ -25,6 +25,9 @@ def configure_api(service, db_password: str, keypair_paths, hostname, is_prod: b
     generate_api_config_file(service, config_path, db_password, hostname,
                              is_prod, orderly_web_api_url, celery_flower_host)
 
+    print("- Making burden estimate files accessible to task queue")
+    service.api.exec_run("chmod a+w tmp/burden_estimate_files")
+
     print("- Sending go signal to API")
     service.api.exec_run("touch {}/go_signal".format(config_path))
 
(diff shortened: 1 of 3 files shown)


src/service_config/api_config.py
```diff
@@ -25,6 +25,9 @@ def configure_api(service, db_password: str, keypair_paths, hostname, is_prod: b
     generate_api_config_file(service, config_path, db_password, hostname,
                              is_prod, orderly_web_api_url, celery_flower_host)
 
+    print("- Making burden estimate files accessible to task queue")
+    service.api.exec_run("chmod a+w tmp/burden_estimate_files")
```
hillalex, line 29: How come this is needed?

hillalex (approved):
Thanks for the careful instructions, works perfectly

src/service_config/api_config.py
EmmaLRussell: I think because the task queue worker runs as a specified user called 'worker'

hillalex: Ah, ok. It's probably not a big deal but this does violate the principle of least privilege, I guess a better solution would be to change the latter to run as root?

EmmaLRussell:
Hmm, so I can get this working with the task queue running as root, with related PR: https://github.com/vimc/montagu-task-queue/pull/28

It does make Celery sad though: `SecurityWarning: You're running the worker with superuser privileges: this is
absolutely not recommended!`

I don't suppose it really matters but it looks like it's going to spam the logs with that warning on every task run. I think I'm inclined to reverting to the worker user and opening up the volume. What do you think?

EmmaLRussell:
> Ugh, yeah that's a pain. There seem to be a few solutions here for this exact scenario that don't involve opening up global write perms on the folder: https://stackoverflow.com/questions/39397548/how-to-give-non-root-user-in-docker-container-access-to-a-volume-mounted-on-the

Thanks for finding that! So I think all of those solutions require the container to know in advance what folder it's going to be working with before it runs (apart from the one where the volume is a mounted host folder, which we're not doing here. I was hoping to avoid having to do that, so that we could use this task to arbitrarily archive any folder, as the task interface suggests. But actually, in practice, since this is the only usage for this task, I think it's OK for the container to know about the `burden_estimate_files` folder in advance.

So this now works with the task queue PR updated with the container worker user doing a `mkdir` on that folder so it gets ownership when the volume is mounted (and not running as root any more).

PR #224, opened 2022-10-03
vimc-6759 updates for new vault

hillalex:
* update vault addr
* update config
* point to updated bb8

Should also merge https://github.com/vimc/bb8/pull/54 (and then can point back to bb8 master here)


weshinsley (approved):
Looks good to me!

